PR #1, opened 2022-08-22
EAHW-2149 Focus on Summary Errors

LaurieBennett:
Added functionality to focus on summary errors when they appear

### DIFF
--- a/src/components/common/form/error-summary/ErrorSummary.jsx
+++ b/src/components/common/form/error-summary/ErrorSummary.jsx
@@ -3,7 +3,6 @@ import { HashLink } from 'react-router-hash-link';
 
 const ErrorSummary = ({ errors }) => {
   const callbackRef = (summaryErrorList) => {
-    // if node exists (there is a summary error), scroll to the first summary error and focus it
     if (summaryErrorList) {
       const firstSummaryError =
         summaryErrorList.children['summary-error-0'].children[
--- a/src/components/common/form/error-summary/ErrorSummary.jsx
+++ b/src/components/common/form/error-summary/ErrorSummary.jsx
@@ -5,9 +5,8 @@ import { HashLink } from 'react-router-hash-link';
 const ErrorSummary = ({ errors }) => {
 
   const callbackRef = (node) => {
-    console.log('Attached node', node)
+    // if node exists (there is a summary error), scroll to the first summary error and focus it
     if(node) {
-      console.log('Scrolling and focusing', node.children["summary-error-0"].children["summary-error-0-message"])
       node.children["summary-error-0"].children["summary-error-0-message"]
         .scrollIntoView({block: "center", inline: "center"});
       node.children["summary-error-0"].children["summary-error-0-message"].focus();
--- a/src/components/common/form/error-summary/ErrorSummary.jsx
+++ b/src/components/common/form/error-summary/ErrorSummary.jsx
@@ -2,15 +2,12 @@ import PropTypes from 'prop-types';
 import { HashLink } from 'react-router-hash-link';
 
 const ErrorSummary = ({ errors }) => {
-  const callbackRef = (node) => {
+  const callbackRef = (summaryErrorList) => {
     // if node exists (there is a summary error), scroll to the first summary error and focus it
-    if (node) {
-      node.children['summary-error-0'].children[
-        'summary-error-0-message'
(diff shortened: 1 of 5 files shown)


src/components/common/form/error-summary/ErrorSummary.jsx
```diff
@@ -2,6 +2,18 @@ import PropTypes from 'prop-types';
 import { HashLink } from 'react-router-hash-link';
 
 const ErrorSummary = ({ errors }) => {
+  const callbackRef = (node) => {
+    // if node exists (there is a summary error), scroll to the first summary error and focus it
+    if (node) {
+      node.children['summary-error-0'].children[
+        'summary-error-0-message'
```
joannafawl, line 9: Could this be put into a variable so that we don't need the comment? Maybe something like `firstSummaryError`?

src/components/common/form/error-summary/ErrorSummary.jsx
```diff
@@ -2,6 +2,18 @@ import PropTypes from 'prop-types';
 import { HashLink } from 'react-router-hash-link';
 
 const ErrorSummary = ({ errors }) => {
+  const callbackRef = (node) => {
+    // if node exists (there is a summary error), scroll to the first summary error and focus it
+    if (node) {
```
joannafawl, line 7: Same with `node` - could that be renamed to something like `summaryError`?

src/components/common/form/error-summary/ErrorSummary.jsx
LaurieBennett: Good call, I've changed the element selection to a separate variable

src/components/common/form/error-summary/ErrorSummary.jsx
LaurieBennett: I've renamed this variable to summaryErrorsList now

joannafawl (dismissed):

src/components/common/form/error-summary/ErrorSummary.jsx
```diff
@@ -2,6 +2,22 @@ import PropTypes from 'prop-types';
 import { HashLink } from 'react-router-hash-link';
 
 const ErrorSummary = ({ errors }) => {
+  const callbackRef = (summaryErrorList) => {
+    // if node exists (there is a summary error), scroll to the first summary error and focus it
```
holly-blues, line 6: I think this comment can be removed

src/components/common/form/error-summary/ErrorSummary.jsx
```diff
@@ -2,6 +2,22 @@ import PropTypes from 'prop-types';
...
+    if (summaryErrorList) {
+      const firstSummaryError =
+        summaryErrorList.children['summary-error-0'].children[
```
holly-blues, line 9: Looks as though `summary-error-` is repeated here and in the "html" below, so might be tempted to add that to a variable

Just a thought, could children dot children be something like the below to make it a little clearer:
`getElementById(variableName).firstChild()`

src/components/common/form/error-summary/ErrorSummary.spec.jsx
```diff
@@ -4,6 +4,7 @@ import { renderWithRouter } from '../../../../test/helpers/Helpers';
 import ErrorSummary from './ErrorSummary';
```
holly-blues, line 5: Is it worth adding a test for checking the focus?

holly-blues (commented):
Like it!
One small concern, I'm hoping the keyboard navigation still works (since the focus does get set) but might be worth checking that it doesn't alter the tab order.

src/components/common/form/error-summary/ErrorSummary.jsx
LaurieBennett: I thought of doing it by firstChild, but explicitly wanted to call out the child just in case anything chaned in the future and the first child wasn't the first error, I can change it if that would be preferred

src/components/common/form/error-summary/ErrorSummary.spec.jsx
LaurieBennett: I've added to the already existing test to check that the focus has been set

holly-blues (approved):

joannafawl (approved):

PR #3, opened 2022-08-24
[EAHW 2047] - Setup hash routing

sewerynmi:


sewerynmi:
Here is the article with the solution https://www.keycloak.org/docs/latest/securing_apps/index.html#_javascript_adapter

LaurieBennett (commented):
Code looks good, just need to add the extra line needed in the .env file to the README.md

holly-blues:
Just for my understanding, what was the problem that needed resolving with those Keycloak changes? Was is related to the Hash Routing change, or something else?

sewerynmi:
@holly-blues With each page reload, the Keycloak was adding 'state' param to the  URL and the APP does not have a Route for that sort of URL. This caused React to not load any component. My solution still have Keycloak 'check user login' functionality, but is happening behind the scenes without using URL's.

holly-blues:
@sewerynmi ok, got it - thanks.

holly-blues (approved):

LaurieBennett (approved):

PR #4, opened 2022-08-25
EAHW-2149 Fix Error Ordering

LaurieBennett:
When a user inputs a valid start time and and invalid finish time, then changes the start time to be an invalid one, the invalid start time error will appear below the invalid finish time error, this is an accessibility issue and is fixed in this pull request.


src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
```diff
@@ -29,22 +30,34 @@ const StartFinishTimeInput = ({
...
   return (
     <div
       className={`govuk-form-group ${
         errorMessages.length > 0 && 'govuk-form-group--error'
       }`}
     >
       <div className="govuk-grid-row">
-        {errorMessages &&
-          errorMessages.map((error, i) => (
+        {sortErrors() &&
```
daleelR, line 52: I think you can get away with just  `sortErrors().map` and remove the first `sortErrors()` check as  `sortErrors().map` should just not render anything if the array is empty.

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
```diff
@@ -29,22 +30,34 @@ const StartFinishTimeInput = ({
     setErrorMessages(relevantErrorMessages);
   };
 
+  const sortErrors = () => {
```
daleelR, line 33: Will all other areas of the ui need to follow this same logic for sorting errors? 
If so maybe it would be worth pulling out into a more generic util function or something that takes an array of the error keys and then call it anywhere that you need to sort the errors.

src/pages/timecard/Timecard.jsx
```diff
@@ -20,11 +20,26 @@ const Timecard = () => {
     document.title = generateDocumentTitle('Timecard');
   });
 
+  const sortSummaryErrors = () => {
```
daleelR, line 23: I guess this is an example of my previous comment, looks like we're duplicating the same kind of logic.

daleelR, line 23: Given it's been identified as a bug that it's possible to get out of order errors, should we have some sort of test to verify that we're always getting sorted errors.

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
LaurieBennett: I've removed the extra check for sortErrors

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
LaurieBennett: I've moved this function now to a separate class

src/pages/timecard/Timecard.jsx
LaurieBennett: As above, I've now moved this function to a separate class

LaurieBennett: I've now added some tests to cover the new functionality

src/utils/sort-errors/sortErrors.js
```diff
@@ -0,0 +1,12 @@
+export const sortErrors = (errors, desiredOrder) => {
+  var orderedKeys = [];
+  const errorKeys = Object.keys(errors);
+  desiredOrder.forEach((key) => {
+    errorKeys.forEach((errorKey) => {
```
daleelR, line 5: This is me just nit picking here really, but I think a nested loop with what looks like n^2 complexity can be avoided here if we were to just convert the errorKeys to a set and then do a errorKeys.has() with the desiredOrder key removing the need for the nested loop. 

In reality it's unlikely to make much of a performance difference with the small number of errors you'd expect to have. 

So i'm not asking for a change or anything just thought it was worth noting.

daleelR (approved):

sureshmanga1 (dismissed):

src/utils/sort-errors/sortErrors.js
LaurieBennett: I've done this now

sureshmanga1 (approved):

daleelR (approved):
Approving as i'm happy with the changes made based on my comments. But expect the outstanding comments from others to be resolved before merge.

src/pages/timecard/Timecard.jsx
```diff
@@ -24,7 +28,7 @@ const Timecard = () => {
     <>
       <BackLink text="Back to calendar" link="/calendar" />
       {summaryErrors && Object.keys(summaryErrors).length !== 0 && (
-        <ErrorSummary errors={summaryErrors} />
+        <ErrorSummary errors={summaryErrors} keys={sortErrors(summaryErrors, desiredErrorOrder)} />
```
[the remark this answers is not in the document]
LaurieBennett: I agree, i've changed the name to sortErrorKeys

src/components/common/form/error-summary/ErrorSummary.jsx
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import { HashLink } from 'react-router-hash-link';
 
-const ErrorSummary = ({ errors }) => {
+const ErrorSummary = ({ errors, keys }) => {
```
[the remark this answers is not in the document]
LaurieBennett: I think this would be desired, but requires complex logic, we should tackle this in the future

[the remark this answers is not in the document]
daleelR: I think a potential way to do it would be to have that sort method sort the actual errors from the error object into an ordered array of errors. E.g. `[{error1, msg1},   {error2, msg2},    {error3 msg3}]` something along those lines.
Then feed that into the error summary object and just update how we extract it. Just an idea.

Perhaps we can create another ticket to cover tackling that and try and implment it early on before more components start getting built. @holly-blues Thoughts?

daleelR (approved):

src/components/common/form/error-summary/ErrorSummary.jsx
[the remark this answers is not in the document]
holly-blues: @daleelR like that idea 
@LaurieBennett agree we have more work to do on error handling, so can get our heads together and tackle this one at that point

holly-blues (approved):

PR #8, opened 2022-08-31
Eahw 2160 get timecard period type

taomoh:
Change the UI to retrieve Timecard Period Types from Time Card API, instead of hard-coded value


src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -24,6 +16,17 @@ const SelectTimecardPeriodType = () => {
...
 
+  const [timePeriods, setTimePeriods] = useState([]);
+  const getResponse = async () => {
+    const response = await getTimePeriodTypes({
+      tenantId: '00000000-0000-0000-0000-000000000000',
+    });
+    setTimePeriods(response.data.items.map((item) => item.name));
+  };
+  useEffect(() => {
+    getResponse();
```
holly-blues, line 27: Name here is a little vague, perhaps getTimePeriodTypeData?

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -24,6 +16,17 @@ const SelectTimecardPeriodType = () => {
     shouldFocusError: false,
   });
 
+  const [timePeriods, setTimePeriods] = useState([]);
+  const getResponse = async () => {
+    const response = await getTimePeriodTypes({
```
holly-blues, line 21: Try/catch here as getTimePeriodTypes throws an error

In case of error getting the data, might we want to set the list of time period types to a default option?

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -24,6 +16,17 @@ const SelectTimecardPeriodType = () => {
...
+      tenantId: '00000000-0000-0000-0000-000000000000',
+    });
+    setTimePeriods(response.data.items.map((item) => item.name));
```
holly-blues, line 24: Perhaps check if response.data.items exist

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -24,6 +16,17 @@ const SelectTimecardPeriodType = () => {
...
+  };
+  useEffect(() => {
+    getResponse();
+  }, []);
```
holly-blues, line 28: This will get the data on every render, which I think is fine for now. Eventually we can look at getting this 'reference data' only once when the application starts and passing the data down to the component

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
```diff
@@ -7,19 +7,55 @@ import SelectTimecardPeriodType from './SelectTimecardPeriodType';
...
-    ];
+jest.mock('../../../api/services/timecardService', () => ({
+  getTimePeriodTypes: () => mockResponse,
+}));
 
+describe('SelectTimecardPeriodType', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('should render a radios component with the correct time periods', async () => {
```
holly-blues, line 58: Shouldn't it also check that the value of the input is set too e.g. to 00000000-0000-0000-0000-000000000001

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
```diff
@@ -42,6 +80,9 @@ describe('SelectTimecardPeriodType', () => {
       />
     );
 
+    // Wait a little bit longer
+    await waitFor(() => {});
```
holly-blues, line 84: Not sure what are we waiting for here...

taomoh: Removing this waitFor(), causes the test to fail with the following message:
`Unable to find an element with the text: You must select a time period. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.`
I seems that the retrieval of the error message element is attempted before its rendering is complete

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
taomoh: Agreed. Function renamed

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
taomoh: Added Try/Catch.

Regarding default options, maybe we should discuss and agree a global approach which would apply across the board (e.g. should the defaults be hard-coded, or cached from first retrieval?).
If we decide, for example, to hard-code the default values, timecardService might be the best place, inside the catch block, meaning the get function will always return a defined value. Let me know your thoughts and I can add an entry for this discussion to the Tech Huddle confluence page

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
taomoh: This is related to the conversation above re: default options

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
holly-blues: Agree we need a global approach and will need to discuss further on defaults (possibly with business too)

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
holly-blues: ah ok - I thought you would usually waitFor('something'). Will have a look into it

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
taomoh: I've added an entry to the Tech Huddle page: https://collaboration.homeoffice.gov.uk/display/EAHW/Tech+Huddle

src/api/services/timecardService.spec.js
```diff
@@ -17,6 +17,17 @@ const timecard = [
...
+  meta: {
+    next: null,
+  },
+  items: [
+    {
+      name: 'Scheduled rest day',
+    },
+  ],
+};
+
```
joannafawl, line 30: Could you import `timeCardPeriodTypes` from the mock data file and test that instead of creating a new variable here?

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
```diff
@@ -1,25 +1,60 @@
...
+      },
+      {
+        name: 'Training',
+      },
+      {
+        name: 'Overtime',
+      },
+    ],
+  },
+};
```
joannafawl, line 38: Same here for importing `timeCardPeriodTypes` from mock data

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -24,6 +16,23 @@ const SelectTimecardPeriodType = () => {
     shouldFocusError: false,
   });
 
+  const [timePeriods, setTimePeriods] = useState([]);
+  const getTimePeriodTypeData = async () => {
+    try {
+      const response = await getTimePeriodTypes({
+        tenantId: '00000000-0000-0000-0000-000000000000',
+      });
```
joannafawl, line 24: The `saveTimeEntry` function in `EditShiftHours` passes the tenant ID in a `URLSearchParams` object, maybe we should keep them consistent?

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
holly-blues: As discussed, I lept ahead here, this is beyond the scope of this requirement/ticket. 
The ticket that tackles a different shift type (e.g. Scheduled Rest Day) will cover checking the input selected (e.g. SRD) has the correct matching value (e.g. ...00002).

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
holly-blues: Change looks good to me.

src/api/services/timecardService.spec.js
taomoh: Good point. Incorporated

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.spec.jsx
taomoh: Good point. Incorporated

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
taomoh: Good point. Incorporated

joannafawl (approved):

holly-blues (approved):

PR #12, opened 2022-09-06
EAHW-2172: Create UI ajax call for getting all time entries

daleelR:
add get time entries ajax call and test. add mock data for get time entries.


LaurieBennett:
Do we not need to do any conversion for the data into the format used in the UI?

daleelR:
> Do we not need to do any conversion for the data into the format used in the UI?

laurieboyes:
Nice thanks for doing this 👍 Do you feel that this would be a good time to add a mock for the endpoint to json server?

daleelR:
> Nice thanks for doing this 👍 Do you feel that this would be a good time to add a mock for the endpoint to json server?

So there's a mock endpoint in `mocks/apiServer.js` there for get already, were you referring to something else?

daleelR:
> Do we not need to do any conversion for the data into the format used in the UI?

@LaurieBennett Have now added

src/utils/time-entry-utils/timeEntryUtils.js
```diff
@@ -0,0 +1,15 @@
+import dayjs from 'dayjs';
+
+export const getSingleTimeEntryResponseItem = (timeEntryResponseData) => {
+  if (timeEntryResponseData.items && timeEntryResponseData.items.length > 0)
+    return timeEntryResponseData.items[0];
+  else return null;
```
LaurieBennett, line 6: Should we throw an error here as well to inform a developer what the issue is?

LaurieBennett (changes requested):

src/utils/time-entry-utils/timeEntryUtils.js
daleelR: Ah yeah I was contemplating this. Think you're right.
I think on the 'else' we should not return anything and just throw an error as then it conforms to the name/purpose of this function.

daleelR: done

LaurieBennett (dismissed):
Looks good, PASS

LaurieBennett (approved):
re-approved

laurieboyes (approved):

PR #15, opened 2022-09-08
EAHW-2172: Display multiple time entries

joannafawl:
- Refactor TimecardContext to use an array for time entries instead of object
- Create function to retrieve time period type by id from database
- Refactor UI to display multiple time entries
- Move timecard date into separate context variable


src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -1,18 +1,18 @@
+import { PropTypes } from 'prop-types';
 import { useState } from 'react';
 import { Link } from 'react-router-dom';
-import { useTimecardContext } from '../../../context/TimecardContext';
 import EditShiftHours from '../edit-shift-hours/EditShiftHours';
 
-const EditShiftTimecard = () => {
+const EditShiftTimecard = ({ timeEntry, index }) => {
```
daleelR, line 6: From the perspective of `editShiftTimecard `I think the variable name `index` isn't too obvious for what the index is referring to (which seems to be the index of the supplied timeEntry within all timeEntries)

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -1,18 +1,18 @@
...
   const toggleEditShiftHours = (event) => {
     event.preventDefault();
     setShowEditShiftHours(!showEditShiftHours);
   };
 
-  const { timecardData } = useTimecardContext();
-  const timecardDataExists = timecardData.startTime !== '';
+  const timeEntriesExists =
```
daleelR, line 12: Should this be named to reflect that a start time exists as it looks like it checks that there is an actual startTime for the given single timeEntry. Rather than mulitple time entries existing, which is what the name suggests.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -6,7 +6,7 @@ import StartFinishTimeInput from '../start-finish-time-input/StartFinishTimeInpu
 import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc';
 
-const EditShiftHours = ({ setShowEditShiftHours }) => {
+const EditShiftHours = ({ setShowEditShiftHours, timeEntry, index }) => {
```
daleelR, line 9: `index` variable naming - same comment as in editShiftTimecard, not completely obvious what it's referring to.

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -48,7 +48,12 @@ const SelectTimecardPeriodType = () => {
       <form
         className="select-timecard-period-type"
         onSubmit={handleSubmit((data) => {
-          setTimecardData({ ...timecardData, timePeriodType: data[radioName] });
+          const newTimeEntries = timeEntries;
+          newTimeEntries.push({
+            timePeriodType: data[radioName],
+            timePeriodTypeId: '00000000-0000-0000-0000-000000000001',
+          });
```
laurieboyes, line 55: Bonus points for making new array in one statement. I did stumble for a minute over the new variable being set to be the same as the old one at first:

```suggestion
          const newTimeEntries [
            ...timeEntries, 
            {
              timePeriodType: data[radioName],
              timePeriodTypeId: '00000000-0000-0000-0000-000000000001',
            }
          ];
```

laurieboyes: Or snap you could even do this inline in the `setTimeEntries()` function and avoid the variable. In my view it would increase the clarity as you don’t have to wonder if the variable is going to be used for any other purpose, but I’ll acknowledge that this is subjective

joannafawl:
> When I refresh the timecard page with existing entries in the DB they aren't shown until you trigger a re-render e.g. select shift type and 'continue'.
> 
> Kinda seems like when the context is updated with the timeEntries from the 'GET' call that it's not triggering a re-render of the Timecard component for some reason. Or maybe something else. Or just something on my end maybe...
> 
> Do you guys get the same result under that condition?

@daleelR have you got the latest version? I think I fixed that with a `Promise.all`

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
joannafawl: Yeah that's much cleaner, just updated now

src/pages/timecard/Timecard.jsx
```diff
@@ -6,20 +6,64 @@ import BackLink from '../../components/common/form/navigation/backlink/BackLink'
...
+    .getUrlSearchParams();
+  const timeEntriesResponse = await getTimeEntries(timeEntriesParams);
+
+  if (timeEntriesResponse.data.items?.length > 0) {
+    const existingTimeEntries = await Promise.all(
+      timeEntriesResponse.data.items.map(async (timeEntry) => {
+        const timePeriodTypeParams = new UrlSearchParamBuilder()
+          .setTenantId('00000000-0000-0000-0000-000000000000')
+          .getUrlSearchParams();
+        const timePeriodType = await getTimePeriodTypeById(
```
laurieboyes, line 36: I think I’m right in saying that this will do a fresh API call for each TimeEntry? I guess this will require a lot of duplicate API calls?

My feeling is that we should aim to minimise the number of API calls our UI has to do where possible, as these will have an impact on the performance of our app and the load on our database.

TimePeriodTypes are likely to change very rarely so I’m thinking we want to cache these responses in some way. _Where_ we cache it is an interesting question. I think in the case of TimePeriodTypes, which will be consistent for a wide range of users, there’s an argument for caching at API level so all users can benefit from the same cache, but that’ll want some thought and is perhaps an optimisation that can come later.

Potentially we get a good effort/value trade-off if we do a bit of caching within the UI, perhaps with a map of cached responses within our function in timecard service?

Interested to get your views and happy to discuss this. Possibly even worth a tech huddle chat 😄

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -28,21 +28,20 @@ const EditShiftHours = ({ setShowEditShiftHours }) => {
...
     const actualStartDateTime = dayjs(
       actualStartDate + ' ' + startTime
     ).format();
 
     const endTime = formData[`${inputName}-finish-time`] || null;
-    let actualEndDateTime;
-    if (endTime) {
-      actualEndDateTime = dayjs(actualStartDate + ' ' + endTime).format();
-    }
+    let actualEndDateTime = endTime
```
laurieboyes, line 42: Nice work changing this to avoid having to reassign the variable on a separate line 👍 
```suggestion
    const actualEndDateTime = endTime
```

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
LaurieBennett: I've changed the index to a more suitable name

src/api/services/timecardService.js
```diff
@@ -36,3 +36,13 @@ export const getTimePeriodTypes = async (params) => {
     );
   }
 };
+
+export const getTimePeriodTypeById = async (id, params) => {
+  try {
+    return await api.get(baseUrl + 'resources/time-period-type/' + id, params);
```
laurieboyes, line 42: Please can you add this endpoint to the mock server?

(side note: Keeping the mock server in sync with reality is going to be a pain. There must be a better way 🤔)

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
LaurieBennett: I've changed the this variable to a more suitable name, it's looking for an empty start date, do determine whether the timecard should show an entry in the edit view

laurieboyes (commented):
Taking me a while so I’ll give you my WIP review 😄

package.json
```diff
@@ -9,8 +9,8 @@
     "prestart:jsonserver": "node mocks/createMockDb.js",
     "start:jsonserver": "node mocks/apiServer.js",
     "serve:mocks": "vite --mode mock",
-    "test": "jest --coverage",
-    "monitor": "jest --watchAll",
+    "test": "TZ=UTC jest --coverage",
```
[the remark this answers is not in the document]
joannafawl: Our drone builds were failing because drone is set to GMT

src/api/services/timecardService.js
sureshmanga1: Yes, This should be configurable as env variable (Not the mock server)

package.json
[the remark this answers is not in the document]
laurieboyes: Cool 👍 ta

src/context/ApplicationContext.jsx
```diff
@@ -0,0 +1,48 @@
...
+    const params = new UrlSearchParamBuilder()
+      .setTenantId('00000000-0000-0000-0000-000000000000')
+      .getUrlSearchParams();
+    const response = await getTimePeriodTypes(params);
+    if (response.data?.items) {
+      return response.data.items;
+    }
+  } catch (error) {
+    console.error(error);
+    return [];
```
laurieboyes, line 19: We’re yet to establish our wider strategy for handling unexpected errors but I’m wondering about this bit.

If we don’t get timeperiodtypes are we able to use the timecard functionality? So is there any value in returning an empty array?

If not, I’d be tempted to stop catching this error and let it bubble up. It’ll make it more obvious that we’ve got work to do to handle random errors 😄

src/context/ApplicationContext.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+const fetchTimePeriodTypesData = async () => {
+  try {
+    const params = new UrlSearchParamBuilder()
+      .setTenantId('00000000-0000-0000-0000-000000000000')
+      .getUrlSearchParams();
+    const response = await getTimePeriodTypes(params);
+    if (response.data?.items) {
+      return response.data.items;
+    }
```
laurieboyes, line 16: (warning: pedantic comment) I think we can make this one line and it’ll have the same behaviour?
```suggestion
    return response.data?.items;
```

src/api/services/timecardService.js
LaurieBennett: I've removed this endpoint and the associated tests as it is no longer used

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
```diff
@@ -48,15 +35,21 @@ const SelectTimecardPeriodType = () => {
       <form
         className="select-timecard-period-type"
         onSubmit={handleSubmit((data) => {
-          setTimecardData({ ...timecardData, timePeriodType: data[radioName] });
+          setTimeEntries([
+            ...timeEntries,
+            {
+              timePeriodType: data[radioName],
+              timePeriodTypeId: '00000000-0000-0000-0000-000000000001',
```
laurieboyes, line 42: Are we still hard coding shift type to shift even though we’re bringing them in from the database now?

src/pages/timecard/Timecard.jsx
```diff
@@ -6,20 +6,53 @@ import BackLink from '../../components/common/form/navigation/backlink/BackLink'
...
+          timeEntryId: timeEntry.id,
+          timePeriodType: timePeriodTypes[timeEntry.timePeriodTypeId],
+          startTime: formatTime(timeEntry.actualStartTime),
+          finishTime: timeEntry.actualEndTime
+            ? formatTime(timeEntry.actualEndTime)
+            : '',
+          timePeriodTypeId: timeEntry.timePeriodTypeId,
+        };
+      })
+    );
```
laurieboyes, line 40: I think I’m right in saying that this doesn’t need to be async anymore?

```suggestion
    const existingTimeEntries = timeEntriesResponse.data.items.map(
      (timeEntry) => ({
        timeEntryId: timeEntry.id,
        timePeriodType: timePeriodTypes[timeEntry.timePeriodTypeId],
        startTime: formatTime(timeEntry.actualStartTime),
        finishTime: timeEntry.actualEndTime
          ? formatTime(timeEntry.actualEndTime)
          : '',
        timePeriodTypeId: timeEntry.timePeriodTypeId,
      })
    );
```

src/api/services/timecardService.spec.js
```diff
@@ -30,112 +32,81 @@ const TENANT_ID_PARAM = {
...
-      expect(response).toBeUndefined();
-    } catch (error) {
-      expect(error).toBeDefined();
-      expect(error.message).toContain('Timecard Service');
-      expect(error.message).toContain('getTimePeriodTypes');
-      expect(error.message).toContain('xyz');
-    }
   });
 
-  describe('getTimeEntries', () => {
```
laurieboyes, line 109: Where the `getTimeEntries` tests removed by mistake?

src/components/timecard/edit-shift-timecard/EditShiftTimecard.spec.jsx
```diff
@@ -8,9 +8,27 @@ import EditShiftTimecard from './EditShiftTimecard';
 
 jest.mock('../../../api/services/timecardService');
 
+let existingTimeEntry;
```
laurieboyes, line 11: Nice refactor 👍

src/pages/timecard/Timecard.jsx
laurieboyes: For the record, we decided to retrieve the TimePeriodTypes just once on app load and make it available via an ‘ApplicationContext’

src/context/ApplicationContext.jsx
LaurieBennett: We added this in as I didn't like the idea of letting the error just happen, but I could add a comment here similar to EditShiftHours with a TODO saying that error handling is to be done when the work is done on handling unexpected errors?

src/pages/timecard/Timecard.jsx
```diff
@@ -66,11 +108,13 @@ const Timecard = () => {
         </Link>
       </div>
 
-      {!timecardData.timePeriodType ? (
-        <SelectTimecardPeriodType />
-      ) : (
-        <EditShiftTimecard />
-      )}
+      {timeEntries.length > 0 &&
```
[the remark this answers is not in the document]
LaurieBennett: I've removed this check now

src/context/ApplicationContext.jsx
LaurieBennett: We can, I've done it

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
LaurieBennett: I've added in some code to retrieve the timePeriodTypeId

src/pages/timecard/Timecard.jsx
LaurieBennett: You're right! Good catch, removed the async

src/api/services/timecardService.spec.js
LaurieBennett: They were, re-added them

src/components/timecard/edit-shift-timecard/EditShiftTimecard.spec.jsx
LaurieBennett: Thanks!

src/pages/timecard/Timecard.jsx
laurieboyes: Ah well there’s that and also that you don’t need the `await` and the `Promise.all`. In my original comment there’s a code suggestion that strips this out.

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
laurieboyes: Lovely 👍 Although I’ve noticed that the unit tests for `SelectTimecardPeriodType` don’t cover the successful submit case. Would you agree that now’s a good time to add that in? Especially now there’s some logic in there for grabbing the selected timeperiodtype

src/pages/timecard/Timecard.jsx
LaurieBennett: I've done this now

src/context/ApplicationContext.jsx
LaurieBennett: As discussed, I've removed this catch statement now

src/components/timecard/select-timecard-period-type/SelectTimecardPeriodType.jsx
LaurieBennett: I've added a test to cover the submit now

laurieboyes: For the record, Laurie explored implementing this test by mocking the `handleSubmit` function that we get back from `useForm` and we agreed that this was super gnarly. We agreed to skip it for now.

laurieboyes (approved):
A thing of beauty. Great work! ✨

daleelR (approved):

PR #16, opened 2022-09-12
Eahw 1353 remove shift time period

LaurieBennett:
Added the ability to remove an existing timecard period with the remove button.

Note: the implementation of the delete modal will be done in a separate PR due to ongoing design discussions

### DIFF
--- a/src/api/services/timecardService.spec.js
+++ b/src/api/services/timecardService.spec.js
@@ -58,7 +58,7 @@ describe('Timecard Service', () => {
       );
     });
 
-    it('should throw useful error containing throwing service and function', async () => {
+    it('should throw useful error containing throwing service and function when an error occurs', async () => {
       api.get.mockImplementation(() => {
         throw new Error('xyz');
       });
@@ -90,7 +90,7 @@ describe('Timecard Service', () => {
       expect(timePeriodTypes.data.items.length).toBe(7);
     });
 
-    it('should throw useful error containing throwing service and function', async () => {
+    it('should throw useful error containing throwing service and function when an error occurs', async () => {
       api.get.mockImplementation(() => {
         throw new Error('xyz');
       });
@@ -129,7 +129,7 @@ describe('Timecard Service', () => {
       expect(response.data).toStrictEqual([newTimeCardEntry]);
     });
 
-    it('should throw a useful error containing throwing service and function', async () => {
+    it('should throw a useful error containing throwing service and function when an error occurs', async () => {
       api.get.mockImplementation(() => {
         throw new Error('xyz');
       });
@@ -159,7 +159,7 @@ describe('Timecard Service', () => {
       expect(response.status).toBeDefined();
     });
 
-    it('should throw a useful error containing throwing service and function', async () => {
+    it('should throw a useful error containing throwing service and function when an error occurs', async () => {
       api.delete.mockImplementation(() => {
         throw new Error('xyz');
       });
--- a/src/api/services/timecardService.spec.js
+++ b/src/api/services/timecardService.spec.js
@@ -1,4 +1,5 @@
 import {
+  deleteTimeEntry,
   getTimecard,
   getTimeEntries,
   getTimePeriodTypes,
@@ -109,6 +110,7 @@ describe('Timecard Service', () => {
       }
     });
   });
+
   describe('getTimeEntries', () => {
     it('should return data correctly on success', async () => {
       api.get.mockImplementation(() =>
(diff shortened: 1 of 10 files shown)


src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -2,19 +2,39 @@ import { PropTypes } from 'prop-types';
...
   const toggleEditShiftHours = (event) => {
     event.preventDefault();
     setShowEditShiftHours(!showEditShiftHours);
   };
 
-  const timeEntryIsEmpty =
+  const { timeEntries, setTimeEntries } = useTimecardContext();
+
+  const timeEntryExists =
     'startTime' in timeEntry && timeEntry.startTime !== '';
```
taomoh, line 19: a shorter and typed version of this expression would be: `timeEntry?.startTime !== ''`

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -2,19 +2,39 @@ import { PropTypes } from 'prop-types';
...
   const [showEditShiftHours, setShowEditShiftHours] = useState(
-    !timeEntryIsEmpty
+    !timeEntryExists
   );
 
+  const clickRemoveShiftButton = async (event) => {
```
daleelR, line 24: This function name sounds like it is automating the action of clicking the remove button.
Might be better named as something a bit more conventional e.g. 'handleClickRemoveShiftButton' or 'handleRemoveShift' etc.

taomoh (commented):
LGTM. Only one little comment

src/api/services/timecardService.spec.js
```diff
@@ -143,4 +145,34 @@ describe('Timecard Service', () => {
...
+      const response = await deleteTimeEntry(123, TENANT_ID_PARAM);
+
+      expect(api.delete).toHaveBeenCalledWith(
+        expect.stringContaining('resources/time-entry/123'),
+        TENANT_ID_PARAM
+      );
+      expect(response.status).toBeDefined();
+    });
+
+    it('should throw a useful error containing throwing service and function', async () => {
```
daleelR, line 162: I think this test name/description should specify a 'when' condition similar to how we specify on the success tests.
e.g '_should throw a useful error containing throwing service and function_ **when an error occurs**'
I can see that it's just a copy paste situation as we've all named the others like that.

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -2,19 +2,39 @@ import { PropTypes } from 'prop-types';
...
+    event.preventDefault();
+    const params = new UrlSearchParamBuilder()
+      .setTenantId('00000000-0000-0000-0000-000000000000')
+      .getUrlSearchParams();
+    const response = await deleteTimeEntry(timeEntry.timeEntryId, params);
+
+    if (response.status === 200) {
```
daleelR, line 31: I think we need some tests to cover this logic here of setting the new time entries that excludes the removed entry.

daleelR:
Was it decided that we weren't going to include implementation for confirming the removal of an entry before actually doing it?
If so, I think the ticket description & comment needs to be updated to reflect that.

LaurieBennett:
> description

Good catch, we'll need to implement this

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
joannafawl: I think when we tried this we had issues because `startTime` doesn't exist in `timeEntry` for new time entries, so it returns true when it should actually return false since it's not a key in the object yet

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
joannafawl: I've updated this

src/api/services/timecardService.spec.js
joannafawl: I've updated this test and the others in the file too

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
joannafawl: We've added a new test for this now

src/api/services/timecardService.spec.js
daleelR: nice one thanks :)

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
daleelR: If it's not too much trouble could we add a similar test to the one you've added / just replace it to cover the situation where you have more than one entry e.g 2 and then you expect `mockSetTimeEntries` to have been called with 1 less entry e.g 1.
Only mentioning as we're using splicing and indexes so it just gives a bit more reassurance. 
If this ends up holding it back then no worries.

joannafawl: Added another one now

daleelR:
- [ ] Update ticket description to mention that the modal work is no longer going to be included as part of this ticket

laurieboyes (approved):
Lovely

daleelR (dismissed):
Approving assuming the outstanding task is completed

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
taomoh: As discussed, the expression I suggested isn't correct. Instead it should have been: 
`const timeEntryExists = timeEntry?.startTime? true : false;`

Or maybe a shorter version to cast to boolean type : `!!timeEntry?.startTime` or `Boolean(timeEntry?.startTime)`

taomoh (approved):

laurieboyes (approved):

PR #17, opened 2022-09-14
[eahw-2171] Use pluaral name in endpoints

sureshmanga1:


laurieboyes (approved):

LaurieBennett (approved):

laurieboyes (dismissed):

joannafawl:
Do you need to update the delete endpoint too?

daleelR (approved):

src/api/services/timecardService.js
```diff
@@ -13,15 +13,15 @@ export const getTimecard = async (params) => {
 
 export const getTimeEntries = async (params) => {
   try {
-    return await api.get(baseUrl + 'resources/time-entry/', params);
+    return await api.get(baseUrl + 'resources/time-entries/', params);
```
taomoh, line 16: This string literal is repeated a few times. It would be better to extract it into a constant

sureshmanga1: I agree this should be configurable value, but we will do it as separate ticket, this should be aligned with server changes.

joannafawl (approved):

PR #18, opened 2022-09-16
EAHW-1348: Scheduled rest day

joannafawl:
#### Changes
- Created `ScheduledRestDay` component
- Updated `Timecard` to render correct time entry component
- Created `removeTimecardContextEntry` util function
- Refactored time entries in context to remove time period type name as it is no longer used
- Refactored scss files
- Added tests


LaurieBennett (changes requested):

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
```diff
@@ -0,0 +1,127 @@
...
+  const handleClickRemoveButton = async (event) => {
+    event.preventDefault();
+    const params = new UrlSearchParamBuilder()
+      .setTenantId('00000000-0000-0000-0000-000000000000')
+      .getUrlSearchParams();
+    const response = await deleteTimeEntry(timeEntry.timeEntryId, params);
+
+    if (response.status === 200) {
+      removeTimecardContextEntry(timeEntries, setTimeEntries, timeEntriesIndex);
+    }
```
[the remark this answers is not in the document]
joannafawl: I copied how we did it in `EditShiftTimecard` to remove shifts as I thought we wanted errors to be handled elsewhere? Might be worth having a conversation soon as a team to decide how to do the error handling

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
```diff
@@ -0,0 +1,127 @@
...
+      .setTenantId('00000000-0000-0000-0000-000000000000')
+      .getUrlSearchParams();
+
+    const response = await createTimeEntry(timecardPayload, params);
+
+    if (response?.data?.items?.length > 0) {
+      const newTimeEntries = deepClone(timeEntries);
+      newTimeEntries[timeEntriesIndex] = ContextTimeEntry.createFrom(timeEntry)
+        .setStartTime(midnight)
+        .setFinishTime(midnight)
```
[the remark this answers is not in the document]
daleelR: This had occurred to me as that's how we retrieve them across separate days for shifts (00:00 - 23:59) however on this Jira ticket it mentions that when the SRD is saved it is recorded as midnight for the start and to midnight the following day for the end. 
So I understood this to be a specific timing to be recorded for SRD.

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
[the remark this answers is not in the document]
daleelR: The general (unimplemented) error handling would've been handled within the timecard service on the actual ajax call. So I think we may not actually need to do any error handling here. 
That does however point to the fact that we're doing a '200' check here which may be redundant as I think the catch block in the TimecardService would prevent us from even reaching this line if there wasn't a 200 response...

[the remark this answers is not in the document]
daleelR: Good idea for the error handling discussion tho as that would clear this up

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
```diff
@@ -0,0 +1,127 @@
...
+    const params = new UrlSearchParamBuilder()
+      .setTenantId('00000000-0000-0000-0000-000000000000')
+      .getUrlSearchParams();
+
+    const response = await createTimeEntry(timecardPayload, params);
+
+    if (response?.data?.items?.length > 0) {
+      const newTimeEntries = deepClone(timeEntries);
+      newTimeEntries[timeEntriesIndex] = ContextTimeEntry.createFrom(timeEntry)
+        .setStartTime(midnight)
```
holly-blues, line 61: shouldn't start and end time come from the successful response? As per TimeEntryId

src/utils/time-entry-utils/timeEntryUtils.js
```diff
@@ -19,3 +22,13 @@ export const formatDate = (dateTime) => {
 export const formatDateTimeISO = (dateTime) => {
   return dayjs(dateTime).format('YYYY-MM-DDTHH:mm:ssZ');
 };
+
+export const removeTimecardContextEntry = (
+  timeEntries,
+  setTimeEntries,
+  removeAtIndex
+) => {
+  const newTimeEntries = deepClone(timeEntries);
```
holly-blues, line 31: I really hope this deepClone is a proper deep copy and not one that doesn't copy functions and other non-serializable objects. If it is the latter it might be time to consider pulling in the use of a library like Lodash vs rolling our own functions and coming up against bugs in the future.

https://developer.mozilla.org/en-US/docs/Glossary/Deep_copy

holly-blues (commented):
It was always the plan to use JSON server to be able to run the UI without needing the backend APIs running.

I don't see any JSON server changes for Scheduled Rest Day and this makes me sad, especially on my first day back at work after a holiday.

Would be good to get this scenario covered, also if there's any better solutions happy to hear them too.

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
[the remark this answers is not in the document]
joannafawl: Resolving this comment with the assumption that we will agree on an approach to handle errors in the UI in tech huddle next week

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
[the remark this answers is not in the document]
joannafawl: Resolving this thread as SRD is saved as 00:00 to 00:00 the next day in the database

src/assets/Timecard.scss
```diff
@@ -1,3 +1,8 @@
+.grey-border {
```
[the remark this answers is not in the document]
joannafawl: We've moved it into a `common.scss` file and used the GDS variables, but we aren't sure about separating the border and padding out - maybe if we need to use a different value for the padding in future it would make sense to split them out then?

src/components/timecard/scheduled-rest-day/ScheduledRestDay.jsx
joannafawl: Updated this now

src/utils/time-entry-utils/timeEntryUtils.js
joannafawl: At the moment, `deepClone` is only use to copy objects or arrays, not functions. Do you think we need to rename the function to reflect that, or would you prefer us to use Lodash instead?

daleelR: Lodash actually looks pretty useful as it has lots of other helpful functions when needing to reconstruct data etc. So i'm all for using Lodash unless we're content with what we have now.

holly-blues (dismissed):

holly-blues (approved):

holly-blues (approved):

mocks/mock-utils.js
```diff
@@ -0,0 +1,41 @@
...
+    ownerId: 1,
+    timePeriodTypeId: '',
+    shiftType: '',
+    actualStartTime: '',
+    actualEndTime: '',
+    plannedStartTime: null,
+    plannedEndTime: null,
+    createdAt: '2022-08-24T12:01:43.786+00:00',
+    updatedAt: '2022-08-24T12:01:43.786+00:00',
+    deleted: false,
```
sureshmanga1, line 24: The following fields are being used and you can remove the other fields. The other fields are required when we start implementing Accrual
id;
tenantId;
ownerId;
timePeriodTypeId;
timePeriodType;
shiftType;
actualStartTime;
actualEndTime;
createdAt;
updatedAt;

mocks/apiServer.js
```diff
@@ -37,32 +39,45 @@ server.use((req, res, next) => {
 });
 
 server.get('/resources/time-entries', function (req, res) {
```
sureshmanga1, line 41: All endpoints should be configurable in one place

sureshmanga1 (approved):

mocks/apiServer.js
joannafawl: Agreed to discuss this in next tech huddle and open a separate ticket for it

src/pages/timecard/Timecard.jsx
```diff
@@ -115,12 +77,54 @@ const Timecard = () => {
...
+    setTimeEntries(existingTimeEntries);
+  } else {
+    setTimeEntries([]);
+  }
+};
+
+const getTimePeriodTypesMap = (timePeriodTypes) => {
+  let timePeriodTypesMap = {};
+  timePeriodTypes.forEach((type) => (timePeriodTypesMap[type.id] = type.name));
+  return timePeriodTypesMap;
```
laurieboyes, line 127:
```suggestion
  return timePeriodTypes.reduce((acc, type) => acc[type.id] = type.name, {});
```

PR #21, opened 2022-09-29
Eahw 1349 add non working day

LaurieBennett:
Added support for Non-working days
Changed Scheduled Rest Day to more generic SimpleTimePeriod component

### DIFF
--- a/src/components/timecard/simple-time-period/SimpleTimePeriod.spec.jsx
+++ b/src/components/timecard/simple-time-period/SimpleTimePeriod.spec.jsx
@@ -41,7 +41,11 @@ jest.mock('../../../api/services/timecardService', () => ({
 describe('SimpleTimePeriod', () => {
   it('should display the correct time period type', () => {
     renderWithTimecardContext(
-      <SimpleTimePeriod timeEntry={newTimeEntry} timeEntriesIndex={0} timePeriodTitle='Scheduled rest day'/>
+      <SimpleTimePeriod
+        timeEntry={newTimeEntry}
+        timeEntriesIndex={0}
+        timePeriodTitle="Scheduled rest day"
+      />
     );
 
     const title = screen.getByText('Scheduled rest day');
@@ -50,7 +54,11 @@ describe('SimpleTimePeriod', () => {
 
   it('should display the save and cancel buttons when there is no time entry', () => {
     renderWithTimecardContext(
-      <SimpleTimePeriod timeEntry={newTimeEntry} timeEntriesIndex={0} timePeriodTitle='Scheduled rest day'/>
+      <SimpleTimePeriod
+        timeEntry={newTimeEntry}
+        timeEntriesIndex={0}
+        timePeriodTitle="Scheduled rest day"
+      />
     );
 
     const saveButton = screen.getByText('Save');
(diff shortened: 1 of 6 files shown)


daleelR:
Mock data needs to be added to support NWD. Should be able to just follow the same way SRD is done

src/components/timecard/simple-time-period/SimpleTimePeriod.spec.jsx
```diff
@@ -38,10 +38,14 @@ jest.mock('../../../api/services/timecardService', () => ({
...
 
-describe('ScheduledRestDay', () => {
+describe('SimpleTimePeriod', () => {
   it('should display the correct time period type', () => {
     renderWithTimecardContext(
-      <ScheduledRestDay timeEntry={newTimeEntry} timeEntriesIndex={0} />
+      <SimpleTimePeriod
+        timeEntry={newTimeEntry}
+        timeEntriesIndex={0}
+        timePeriodTitle="Scheduled rest day"
```
daleelR, line 47: nit: I think this title param could have been extracted out to the top level as its used all across this test suite. 
Not necessary to change though

src/pages/timecard/Timecard.spec.jsx
```diff
@@ -82,9 +82,32 @@ describe('Timecard', () => {
 
     expect(screen.queryByText('Add a new time period')).toBeFalsy();
     expect(screen.queryByText('Shift')).toBeFalsy();
+    expect(screen.queryByText('Non-working day')).toBeFalsy();
     expect(screen.getByText('Scheduled rest day')).toBeTruthy();
   });
 
+  it('should render the NonWorkingDay component when time period type is NWD', async () => {
```
daleelR, line 89: The test name needs to change to refer to the SimpleTimePeriod component. Same goes for the SRD one on the test above.
```suggestion
  it('should render the SimpleTimePeriod component when time period type is NWD', async () => {
```

src/pages/timecard/Timecard.spec.jsx
```diff
@@ -82,9 +82,32 @@ describe('Timecard', () => {
...
+      ],
+      setTimeEntries: jest.fn(),
+      timecardDate: '',
+      setTimecardDate: jest.fn(),
+    });
+
+    expect(screen.queryByText('Add a new time period')).toBeFalsy();
+    expect(screen.queryByText('Shift')).toBeFalsy();
+    expect(screen.queryByText('Scheduled rest day')).toBeFalsy();
+    expect(screen.getByText('Non-working day')).toBeTruthy();
```
daleelR, line 108: Since we're implementing more time period types now we'll have to keep remembering to add each type to these tests. 
It might be better to loop though the time periods and test them all which will keep the tests up to date when we implement more periods.

Could do something like this if you import the time periods from the mock data and then apply to the shift, SRD and this test:
```suggestion
  const testSimpleTimePeriodRendersWithCorrectTitle = (timePeriodTypeTitle) => {
    expect(screen.getByText(timePeriodTypeTitle)).toBeTruthy();
    timeCardPeriodTypes.forEach(type => {
      if (type.name != timePeriodTypeTitle) expect(screen.queryByText(type.name)).toBeFalsy();
  })
}
```

LaurieBennett: Done

src/pages/timecard/Timecard.spec.jsx
LaurieBennett: I've changed this from being an individual test to a test.each, so the title has changed now

daleelR: perfect.

daleelR (approved):

JWeston-HO (approved):

PR #28, opened 2022-10-12
Remove hint text

JWeston-HO:
Following on from a review it was decided the hint text should be removed from add a time entry. This PR removes that hint text. 

Jira Ticket: https://collaboration.homeoffice.gov.uk/jira/browse/EAHW-2244

### DIFF
--- a/src/components/timecard/add-timecard-period/AddTimeCardPeriod.jsx
+++ b/src/components/timecard/add-timecard-period/AddTimeCardPeriod.jsx
@@ -6,9 +6,6 @@ const AddTimeCardPeriod = ({ timecardEmpty }) => {
   const addPeriodTitle = timecardEmpty
     ? 'Add time period'
     : 'Add another time period';
-  const addPeriodDescription = timecardEmpty
-    ? 'Use this to record a shift'
-    : 'Use this to record overtime or another shift';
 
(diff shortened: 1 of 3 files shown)


src/components/timecard/add-timecard-period/AddTimeCardPeriod.jsx
```diff
@@ -36,9 +33,6 @@ const AddTimeCardPeriod = ({ timecardEmpty }) => {
               </Link>
             </div>
           </div>
-          <p className="govuk-body-m govuk-secondary-text-colour">
```
joannafawl, line 39: The `govuk-secondary-text-colour` style isn't being used anywhere anymore, do we still want to keep it? Happy to leave it if we think we will use it elsewhere

holly-blues (commented):
Presume after the change it has adjusted the alignment automatically?

src/components/timecard/add-timecard-period/AddTimeCardPeriod.jsx
JWeston-HO: removed from common.scss

holly-blues (approved):

joannafawl (approved):

PR #29, opened 2022-10-13
EAHW-2226 initial commit

LaurieBennett:
- Retrieve the PersonID from keycloak and map it to the userID in the application context
- Retrieve the user's timecard claim from keycloak and map it to the timecardClaim in the application conext
- Update the uses of userID in the ui to use the newly mapped variable

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -38,6 +38,19 @@ steps:
     depends_on:
       - install dependencies
 
+  - name: sonar analysis
+    image: *node_image
+    environment:
+      SONAR_TOKEN:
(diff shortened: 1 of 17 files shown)


JWeston-HO (dismissed):

daleelR:
This needs to be updated to work in mock mode. I think it's not retrieving the userId in mock mode causing undefined to pass in the request failing it.

- [ ] Fix mock mode

taomoh (approved):

daleelR (approved):

PR #32, opened 2022-10-21
EAHW-928: Validate time entry clashes

joannafawl:
Added functionality for displaying error messages when adding or updating a shift that clashes with another time period. 

If running locally, you will need to use the `feature/eahw-928/time-entry-of-time-period-validation-on-save` branch of the timecard api.

### Changes
- Remove try / catch block from `createTimeEntry` and `updateTimeEntry` to allow errors to be caught and handled by `validateServiceErrors`
- Use `handleServerValidationErrors` function to catch server errors and display relevant error messages for clashes (note: this will change in the next ticket EAHW-2239)
- Add validation logic to mock data
- Add tests

Screenshot 2022-10-25 at 16 55 24

### DIFF
--- a/mocks/apiServer.js
+++ b/mocks/apiServer.js
@@ -1,4 +1,7 @@
 /* eslint-disable no-console */
+const isSameOrBefore = require('dayjs/plugin/isSameOrBefore');
+const isSameOrAfter = require('dayjs/plugin/isSameOrAfter');
+const dayjs = require('dayjs');
 const jsonServer = require('json-server');
 const server = jsonServer.create();
 const path = require('path');
@@ -108,5 +111,28 @@ server.listen(port, () => {
 
 function validateTimecard(timecard) {
   if (!timecard.actualStartTime) return 'Start time is required.';
+
+  dayjs.extend(isSameOrBefore);
+  dayjs.extend(isSameOrAfter);
+  for (existingTimeEntryKey in timePeriodIdForTimeEntry) {
+    const newStartTime = dayjs(timecard.actualStartTime);
+    const existingStartTime = dayjs(
+      timePeriodIdForTimeEntry[existingTimeEntryKey].actualStartTime
+    );
+    const existingEndTime = dayjs(
+      timePeriodIdForTimeEntry[existingTimeEntryKey].actualEndTime
+    );
+
+    if (
+      newStartTime.isSameOrAfter(existingStartTime) &&
+      newStartTime.isSameOrBefore(existingEndTime)
+    ) {
+      return {
+        message:
+          ' has the following error(s): Time periods must not overlap with another time period',
+      };
+    }
+  }
+
   return '';
 }
--- a/mocks/mockData.js
+++ b/mocks/mockData.js
@@ -1,4 +1,5 @@
 const { createTimeEntry } = require('./mock-utils');
+const dayjs = require('dayjs');
 
 const accruals = [
   {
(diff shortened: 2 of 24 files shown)


src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -66,43 +96,51 @@ const EditShiftHours = ({
...
+            .setStartTime(formattedStartTime)
+            .setFinishTime(formattedEndTime)
+            .setTimeEntryId(responseItem.id);
+
+          setTimeEntries(newTimeEntries);
+          setSummaryErrors({});
+          setShowEditShiftHours(false);
+        }
+      },
+      handleServerValidationErrors
```
LaurieBennett, line 134: Shouldn't this be "handleServerValidationErrors()"?

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -37,14 +37,44 @@ const EditShiftHours = ({
...
+      summaryErrors['shift-finish-time'] = {
+        message: '',
+      };
+    } else {
+      errorsHandled = false;
+    }
+
+    if (errorsHandled) {
+      setSummaryErrors(summaryErrors);
+      return true;
```
LaurieBennett, line 73: What are these returned values being used for? Can we get rid of them?

LaurieBennett (changes requested):

src/pages/timecard/Timecard.jsx
```diff
@@ -157,6 +157,7 @@ const updateTimeEntryContextData = async (
         setTimeEntries([]);
       }
     },
+    () => {},
```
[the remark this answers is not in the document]
joannafawl: We need to pass in a function for the third argument in order to set the `isRecoverable` boolean, but right now we don't have any custom errors in `Timecard` so it's just an empty function. It's not ideal but it's the best we could come up with, open to any suggestions though :)

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
joannafawl: We are just passing in a reference here so that it can be used inside `validateServiceErrors`

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
joannafawl: This is a tricky one because ideally this function would only be setting error messages, but in the situation where a backend error is thrown and we don't have a custom error message for it, we would want to show the service error page which is determined in `validateServiceErrors`. The only way we could get the information on whether all the errors have been handled is by returning a boolean

mocks/apiServer.js
```diff
@@ -108,5 +111,28 @@ server.listen(port, () => {
 
 function validateTimecard(timecard) {
   if (!timecard.actualStartTime) return 'Start time is required.';
+
+  dayjs.extend(isSameOrBefore);
```
laurieboyes, line 115: I love that you’ve built this into the mock server! 😄

mocks/mockData.js
```diff
@@ -183,37 +184,53 @@ const people = [
...
-const userId = "9e47cf29-1598-4269-aa31-db1d2e4e0207";
+const userId = '9e47cf29-1598-4269-aa31-db1d2e4e0207';
 
 const shiftTimeEntry = createTimeEntry({
   id: 'c0a80040-82cf-1986-8182-cfedbbd50003',
   timePeriodTypeId: '00000000-0000-0000-0000-000000000001',
-  actualStartTime: '2022-08-24T12:01:43.786+00:00',
-  actualEndTime: '2022-08-24T22:01:43.786+00:00',
+  actualStartTime: dayjs().hour(12).minute(0).format('YYYY-MM-DDTHH:mm:ssZ'),
+  actualEndTime: dayjs().hour(22).minute(0).format('YYYY-MM-DDTHH:mm:ssZ'),
```
laurieboyes, line 193: I’m interested in the dayjs usage here. How come you get away with not setting a day? Does it default to today’s date? I guess that’s pretty neat. Don’t want to have to be clicking back ages to see mock data. Cool 👍 (if my assumption is correct)

src/api/services/timecardService.js
```diff
@@ -20,23 +20,15 @@ export const getTimeEntries = async (params) => {
...
 };
 
 export const updateTimeEntry = async (id, payload, params) => {
-  try {
-    return await api.put(
-      baseUrl + 'resources/time-entries/' + id,
-      params,
-      payload
-    );
-  } catch (error) {
```
laurieboyes, line 37: Hooray for centralised error handling 🎉

src/utils/api-utils/ApiUtils.js
```diff
@@ -10,6 +11,15 @@ export const validateServiceErrors = async (
     });
   } catch (error) {
     console.error(error);
+    const allErrorsHandled = handleCustomErrors(error?.response?.data?.message);
```
laurieboyes, line 14: I’m a little bothered by not being able to understand what this function does and what it returns from its name. I’ve figured it out though and I do get it, and personally can’t think of a better name. I think this would particularly benefit from TypeScript to document what’s expected of this function (That’s obviously vastly out of scope for this ticket 😄 just thinking aloud).

laurieboyes (commented):
Sorry, review is a work in progress as I need to jump onto something else for a min

src/pages/timecard/Timecard.jsx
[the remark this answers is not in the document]
laurieboyes: Could you assign the empty function to a variable called `handleCustomErrors` first to provide a bit more signposting for why it’s being passed in?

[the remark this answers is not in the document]
joannafawl: Good idea, I've updated it now

mocks/mockData.js
joannafawl: Yeah `dayjs()` defaults to today. The mock data shows on every page no matter the date (which is also how it was working when it was set to 24th August), but before we made this change, we weren't getting any clashes because the dates were different

laurieboyes (approved):

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -37,14 +37,44 @@ const EditShiftHours = ({
...
     setSummaryErrors(errorFields);
   };
 
+  const handleServerValidationErrors = (error) => {
+    const summaryErrors = {};
+    let errorsHandled = true;
+
+    if (
+      error ==
+      ' has the following error(s): Time periods must not overlap with another time period'
```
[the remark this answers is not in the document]
joannafawl: This is temporary - we had some issues when using `includes`, but we are going to change it in eahw-2239 so we didn't want to spend too long on it

PR #39, opened 2022-11-01
Eahw 589 input day after start time

JWeston-HO:
This PR also covers EAHW-1724. 
It contains the logic to add text and date to the screen when the user enters a finish time that is numerically before the start time. It will now assume this  finish time is on the next day and persist data accordingly. 

Added text added to the screen.
 
image

image

### DIFF
--- a/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
+++ b/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
@@ -15,7 +15,6 @@ const ValidatedTimeEntry = ({
   timeEntry,
   timeEntriesIndex,
 }) => {
-
   const { timeEntries, setTimeEntries } = useTimecardContext();
 
   const setFinishTimeText = () => {
--- a/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
+++ b/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
@@ -69,7 +69,6 @@ export default ValidatedTimeEntry;
 ValidatedTimeEntry.propTypes = {
   name: PropTypes.string.isRequired,
   timeType: PropTypes.string.isRequired,
-  timeIndetifier: PropTypes.string.isRequired,
   errors: PropTypes.object.isRequired,
   defaultValue: PropTypes.string,
   register: PropTypes.any.isRequired,
--- a/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
+++ b/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
@@ -18,13 +18,11 @@ const ValidatedTimeEntry = ({
   }, [formState]);
 
   const isFinishTimeNextDay = () => {
-    if (document.getElementById(`${name}-finish-time`)) {
-      var finishTimeValue = document.getElementById(
-        `${name}-finish-time`
-      ).value;
+    if (document.getElementById(`shift-finish-time`)) {
+      var finishTimeValue = document.getElementById(`shift-finish-time`).value;
     }
-    if (document.getElementById(`${name}-start-time`)) {
-      var startTimeValue = document.getElementById(`${name}-start-time`).value;
+    if (document.getElementById(`shift-start-time`)) {
+      var startTimeValue = document.getElementById(`shift-start-time`).value;
     }
     var answer = calculateFinishTimeOnNextDay(startTimeValue, finishTimeValue);
 
@@ -71,6 +69,7 @@ export default ValidatedTimeEntry;
 ValidatedTimeEntry.propTypes = {
   name: PropTypes.string.isRequired,
   timeType: PropTypes.string.isRequired,
+  timeIndetifier: PropTypes.string.isRequired,
   errors: PropTypes.object.isRequired,
   defaultValue: PropTypes.string,
   register: PropTypes.any.isRequired,
--- a/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
+++ b/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
@@ -1,5 +1,4 @@
 import PropTypes from 'prop-types';
-import { useEffect } from 'react';
 import { useTimecardContext } from '../../../../context/TimecardContext';
 import { deepCloneJson } from '../../../../utils/common-utils/common-utils';
 import { ContextTimeEntry } from '../../../../utils/time-entry-utils/ContextTimeEntry';
@@ -12,7 +11,6 @@ const ValidatedTimeEntry = ({
   defaultValue,
   register,
   isRequired,
-  formState,
   getValues,
   timeEntry,
   timeEntriesIndex,
@@ -20,10 +18,6 @@ const ValidatedTimeEntry = ({
   const errorMessage = `You must enter a ${timeType} in the HH:MM 24 hour clock format`;
 
   const { timeEntries, setTimeEntries } = useTimecardContext();
-
-  useEffect(() => {
-    setFinishTimeText();
-  }, [formState]);
 
   const setFinishTimeText = () => {
     const startTimeValue = getValues('shift-start-time');
--- a/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
+++ b/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
@@ -29,7 +29,7 @@ const ValidatedTimeEntry = ({
     if (document.getElementById('end-next-day')) {
       if (answer >= 1) {
         document.getElementById('end-next-day').innerHTML =
-          'Shift ending on next day';
+          'Finishes on next day';
       } else {
         document.getElementById('end-next-day').innerHTML = '';
       }
--- a/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
+++ b/src/components/common/validation/time-format/ValidatedTimeEntry.jsx
@@ -1,6 +1,9 @@
 import PropTypes from 'prop-types';
 import { useEffect } from 'react';
-import { calculateFinishTimeOnNextDay } from '../../../../utils/time-entry-utils/timeEntryUtils';
+import { useTimecardContext } from '../../../../context/TimecardContext';
+import { deepCloneJson } from '../../../../utils/common-utils/common-utils';
+import { ContextTimeEntry } from '../../../../utils/time-entry-utils/ContextTimeEntry';
+import { isFinishTimeOnNextDay } from '../../../../utils/time-entry-utils/timeEntryUtils';
 
 const ValidatedTimeEntry = ({
   name,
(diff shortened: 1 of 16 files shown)


src/components/timecard/edit-shift-hours/EditShiftHours.spec.jsx
```diff
@@ -72,7 +72,7 @@ describe('EditShiftHours', () => {
 
     it('should call updateTimeEntry when pressing save when there is an existing time entry', async () => {
       const timeEntryId = '1';
-      const inputtedEndTime = '06:00';
+      const inputtedEndTime = '10:00';
```
LaurieBennett, line 75: Is there a reason for this test update?

src/utils/time-entry-utils/timeEntryUtils.js
```diff
@@ -32,3 +32,15 @@ export const removeTimecardContextEntry = (
   newTimeEntries.splice(removeAtIndex, 1);
   setTimeEntries(newTimeEntries);
 };
+
+export const calculateFinishTimeOnNextDay = (
+  startTimeValue,
+  finishTimeValue
+) => {
+  if (startTimeValue && finishTimeValue) {
+    return Math.floor(
```
daleelR, line 41: I'm assuming the purpose of this function is to determine if the finish time is on the next day.
It looks like we're doing some sort of manual conversion of the time values here to get that result.
If so, we should make use of the dayJs library to greatly simplify this operation and return a boolean instead.

LaurieBennett (changes requested):
We seem to be calculating whether the finish date is on another day multiple times, can we calculate this once and pass it down instead? I understand the onBlur probably needs to stay though, but can we pass down a method instead?

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -35,6 +35,14 @@ const EditShiftTimecard = ({ timeEntry, timeEntriesIndex }) => {
     });
   };
 
+  // const isFinishTimeNextDay = (startTimeValue, finishTimeValue) => {
```
daleelR, line 38: this function name suggests it's returning a boolean when in fact its giving us the output text.
Could we either change the function name, or move the display text into the jsx.

src/utils/time-entry-utils/ContextTimeEntry.js
```diff
@@ -39,6 +48,11 @@ export class ContextTimeEntry {
     return this;
   }
 
+  setFinishDate(finishDate) {
+    this.finishDate = finishDate;
+    return this;
+  }
```
joannafawl, line 54: Do we need a finish date? `finishTime` is a dateTime, so we previously decided to get rid of it

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -50,6 +51,18 @@ const EditShiftHours = ({
     setSummaryErrors(errorFields);
   };
 
+  const isFinishTimeNextDay = (
```
daleelR, line 54: Simmilar comment here on the function naming. The name suggests a boolean return value, but it's returning a date.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -85,8 +98,18 @@ const EditShiftHours = ({
     );
 
     const endTime = formData[`${inputName}-finish-time`] || null;
+    var actualEndDate = formatDate(timecardDate);
```
daleelR, line 101: we should avoid using `var `where not necessary as it's function scoped. 
might need to use `let `as it's reassigned a value on ln 104

src/components/common/validation/time-format/ValidatedTimeEntry.spec.jsx
```diff
@@ -6,8 +6,8 @@ describe('ValidatedTimeEntry', () => {
   it('should render correctly', () => {
     const component = render(
       <ValidatedTimeEntry
-        name="start-time"
-        timeType="start time"
+        name="shift-finish-time"
```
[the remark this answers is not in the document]
JWeston-HO: Another test has been added

src/utils/time-entry-utils/timeEntryUtils.js
```diff
@@ -37,14 +40,11 @@ export const removeTimecardContextEntry = (
...
-) => {
+export const isFinishTimeOnNextDay = (startTimeValue, finishTimeValue) => {
   if (startTimeValue && finishTimeValue) {
-    return Math.floor(
-      parseFloat(startTimeValue.replace(/:/g, '.')).toFixed(2) /
-        parseFloat(finishTimeValue.replace(/:/g, '.')).toFixed(2)
-    );
+    return dayjs(
+      dayjs().format('YYYY-MM-DD') + '' + finishTimeValue
+    ).isSameOrBefore(dayjs(dayjs().format('YYYY-MM-DD') + '' + startTimeValue));
```
daleelR, line 47: I think this can be further cut down to the following by using the existing formatDate util function

```suggestion
    return dayjs(formatDate(dayjs()) + finishTimeValue).isSameOrBefore(
      dayjs(formatDate(dayjs()) + startTimeValue)
    );
```

src/components/common/validation/time-format/ValidatedTimeEntry.jsx
```diff
@@ -9,32 +11,26 @@ const ValidatedTimeEntry = ({
...
 
-    if (document.getElementById('end-next-day')) {
-      if (answer >= 1) {
-        document.getElementById('end-next-day').innerHTML =
-          'Finishes on next day';
-      } else {
-        document.getElementById('end-next-day').innerHTML = '';
-      }
-    }
+    let answer = isFinishTimeOnNextDay(startTimeValue, finishTimeValue);
```
daleelR, line 27: This should be declared as const as we won't be changing it value, and also needs a meaningful name.
Or we could just pass this function call directly into line 30.

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
```diff
@@ -8,6 +8,9 @@ const StartFinishTimeInput = ({
   errors,
   startTimeValue,
   finishTimeValue,
+  getValues,
```
daleelR, line 11: this `getValues` name is passed down all the way through to the `ValidateTimeEntry` component and from the context of this component and the child ones `getValues` doesn't provide much meaning. 
I think the prop should be named to something that tells us what kind of values it's getting. e.g. `getFormValues` (and apply to child components too).

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -51,12 +51,8 @@ const EditShiftHours = ({
     setSummaryErrors(errorFields);
   };
 
-  const isFinishTimeNextDay = (
-    actualStartTime,
-    actualEndTime,
-    actualStartDate
-  ) => {
-    if (calculateFinishTimeOnNextDay(actualStartTime, actualEndTime) >= 1) {
+  const setFinishTimeDate = (actualStartDate) => {
```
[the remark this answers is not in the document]
JWeston-HO: we are setting the actualEndDate Ln 96. We kinda get the finishNextDay context variable. But use that to set the date to persist.

src/components/common/validation/time-format/ValidatedTimeEntry.jsx
```diff
@@ -42,8 +69,11 @@ export default ValidatedTimeEntry;
 ValidatedTimeEntry.propTypes = {
   name: PropTypes.string.isRequired,
   timeType: PropTypes.string.isRequired,
+  timeIndetifier: PropTypes.string.isRequired,
   errors: PropTypes.object.isRequired,
   defaultValue: PropTypes.string,
   register: PropTypes.any.isRequired,
   isRequired: PropTypes.bool,
+  formState: PropTypes.any,
+  onBlur: PropTypes.bool,
```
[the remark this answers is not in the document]
JWeston-HO: These have been removed

src/components/common/validation/time-format/ValidatedTimeEntry.jsx
```diff
@@ -7,8 +9,32 @@ const ValidatedTimeEntry = ({
...
 }) => {
   const errorMessage = `You must enter a ${timeType} in the HH:MM 24 hour clock format`;
+
+  useEffect(() => {
+    isFinishTimeNextDay();
+  }, [formState]);
+
+  const isFinishTimeNextDay = () => {
+    if (document.getElementById(`shift-finish-time`)) {
+      var finishTimeValue = document.getElementById(`shift-finish-time`).value;
```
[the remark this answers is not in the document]
JWeston-HO: This line has now been removed

src/components/common/validation/time-format/ValidatedTimeEntry.jsx
```diff
@@ -7,8 +9,32 @@ const ValidatedTimeEntry = ({
...
+    }
+    if (document.getElementById(`shift-start-time`)) {
+      var startTimeValue = document.getElementById(`shift-start-time`).value;
```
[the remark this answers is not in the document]
JWeston-HO: This line has now been removed

src/components/timecard/edit-shift-hours/EditShiftHours.spec.jsx
JWeston-HO: Reverted

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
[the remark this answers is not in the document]
daleelR: yeah we're setting a value to `actualEndDate` on ln 96 but that is just a return value from `setFinishTimeDate`.
But here on line 54 the current function name should describe what it's implementation does and currently it's implementation doesn't set anything.
The `actualStartDate` param remains unchanged, and no state is changed either.

src/utils/time-entry-utils/timeEntryUtils.spec.js
```diff
@@ -43,4 +44,34 @@ describe('timeEntryUtils', () => {
     removeTimecardContextEntry(existingTimeEntries, mockSetTimeEntries, 1);
     expect(mockSetTimeEntries).toHaveBeenCalledWith([timeEntry1]);
   });
+
+  describe('isFinishTimeOnNextDay', () => {
+    it('should return true when finish time is before start time', async () => {
+      const startTime = '12:03';
+      const finishTime = '01:01';
```
daleelR, line 51: Can we make this more of a boundary test to test the latest time it can possibly be to be considered the next day. E.g. one minute before start time.
Then maybe another simmilar one which is the same time as the start time which would also return true.
```suggestion
      const finishTime = '12:02';
```

src/utils/time-entry-utils/timeEntryUtils.spec.js
```diff
@@ -43,4 +44,34 @@ describe('timeEntryUtils', () => {
...
+
```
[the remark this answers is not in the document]
JWeston-HO: Add more tests to cover

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
JWeston-HO: updated to getFormValues

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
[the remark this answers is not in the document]
JWeston-HO: changed to get

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
```diff
@@ -91,11 +95,18 @@ const StartFinishTimeInput = ({
...
+              timeEntry={timeEntry}
+              timeEntriesIndex={timeEntriesIndex}
             />
           </div>
         </div>
+        <p
+          id="end-next-day"
+          className="govuk-hint govuk-grid-column-one-third govuk-!-padding-top-8 govuk-!-padding-right-0 govuk-!-padding-left-0"
+        >
+          {timeEntry.finishNextDay ? 'Finishes next day' : ''}
```
daleelR, line 108: Although the ticket description didn't mention any specific text, the mockup suggests the following text:

> Shift ending on next day

Perhaps worth a change..

JWeston-HO: See ticket 1724 - https://collaboration.homeoffice.gov.uk/jira/browse/EAHW-1724
I spoke with Olivia and confirmed her comment on that ticket to be correct.

src/components/common/validation/time-format/ValidatedTimeEntry.jsx
JWeston-HO: updated

src/components/timecard/start-finish-time-input/StartFinishTimeInput.jsx
daleelR: Ah yes, didn't see that comment.

LaurieBennett (approved):

PR #42, opened 2022-11-03
EAHW-2299 Bug: empty shift in timecard

LaurieBennett:
moved setting of setShowEditShiftHors to useEffect to ensu it triggers setter on re-render

### DIFF
--- a/src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
+++ b/src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
@@ -5,7 +5,11 @@ import EditShiftHours from '../edit-shift-hours/EditShiftHours';
 import { deleteTimeEntry } from '../../../api/services/timecardService';
 import { UrlSearchParamBuilder } from '../../../utils/api-utils/UrlSearchParamBuilder';
 import { useTimecardContext } from '../../../context/TimecardContext';
-import { removeTimecardContextEntry } from '../../../utils/time-entry-utils/timeEntryUtils';
+import {
+  formatDateNoYear,
+  formatTime,
+  removeTimecardContextEntry,
+} from '../../../utils/time-entry-utils/timeEntryUtils';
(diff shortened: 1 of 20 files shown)


claire-guest (approved):

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -14,19 +14,25 @@ import { validateServiceErrors } from '../../../utils/api-utils/ApiUtils';
...
-  const toggleEditShiftHours = (event) => {
-    event.preventDefault();
-    setShowEditShiftHours(!showEditShiftHours);
-  };
-
   const { setServiceError } = useApplicationContext();
   const { timeEntries, setTimeEntries } = useTimecardContext();
 
-  const timeEntryExists = !!timeEntry?.startTime;
+  const timeEntryExists = !!timeEntry?.startTime && timeEntry?.startTime !== '';
```
daleelR, line 20: nit: I don't think you need the second question mark `timeEntry?` as you know you have the object from the first condition.
Think it can be: `...&& timeEntry.startTime !== ''`

daleelR (commented):
Is it possible to have a corresponding test to assert that the previous bugged behavior is no longer possible.

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
LaurieBennett: Done

LaurieBennett:
> Is it possible to have a corresponding test to assert that the previous bugged behavior is no longer possible?

I haven't been able to write one that will consistently work, this behaviour will also stop being possible when the confirmation modal is added, as the user will be unable to click away from the in progress add timecard entry without being told their unsaved timecard will be lost.

daleelR (approved):

claire-guest (approved):

PR #45, opened 2022-11-09
Eahw 2239/show clashes in errors

claire-guest:
Add details of clashing shifts into the displayed error. Only the error by the entry boxes displays the clashing errors, whilst the summary box at the top of the page displays the generic error.
Screenshot 2022-11-09 at 15 25 45
Screenshot 2022-11-09 at 15 26 04

### DIFF
--- a/src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
+++ b/src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
@@ -7,7 +7,6 @@ const NotesSummaryList = () => {
   const [addNote, setAddNote] = useState(false);
 
   const handleClick = () => {
-    event.preventDefault();
     setAddNote(!addNote);
   };
 
--- a/src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
+++ b/src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
@@ -7,6 +7,7 @@ const NotesSummaryList = () => {
   const [addNote, setAddNote] = useState(false);
 
   const handleClick = () => {
+    event.preventDefault();
     setAddNote(!addNote);
   };
 
--- a/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
+++ b/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
@@ -12,11 +12,9 @@ import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc';
 import { UrlSearchParamBuilder } from '../../../utils/api-utils/UrlSearchParamBuilder';
 import {
-  ClashingProperty,
   formatDate,
   formatDateTimeISO,
   formatTime,
-  inputNames,
 } from '../../../utils/time-entry-utils/timeEntryUtils';
 import { ContextTimeEntry } from '../../../utils/time-entry-utils/ContextTimeEntry';
 import {
@@ -27,6 +25,7 @@ import { validateServiceErrors } from '../../../utils/api-utils/ApiUtils';
 import { useState } from 'react';
 import TimeInputErrors from '../time-input-errors/TimeInputErrors';
 import { useEffect } from 'react';
+import { ClashingProperty, inputNames } from '../../../utils/constants';
 
 const EditShiftHours = ({
   setShowEditShiftHours,
--- a/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
+++ b/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
@@ -45,10 +45,52 @@ const EditShiftHours = ({
     summaryErrors,
     setSummaryErrors,
   } = useTimecardContext();
-  const [wholeShiftClash, setWholeShiftClash] = useState(false);
+  const [clashingProperty, setClashingProperty] = useState(null);
+  const [clashingTimes, setClashingTimes] = useState(null);
 
   const handleError = (errorFields) => {
     setSummaryErrors(errorFields);
+  };
+
+  const formatTimeClashes = (timeClashes) => {
+    if (timeClashes.length > 1) {
+      var times = (
+        <div>
+          <p>You are already assigned to the following time periods:</p>
+          <ul>
+            {timeClashes.map((clash) => (
+              <li key="foo">{shiftClashToText(clash)}</li>
+            ))}
+          </ul>
+        </div>
+      );
+      return times;
+    } else {
+      var clash = timeClashes[0];
+      var startTime = Date.parse(clash.startTime);
+      var endTime = Date.parse(clash.endTime);
+      var text;
+      if (clash.timePeriodTypeId == '00000000-0000-0000-0000-000000000001') {
+        text = `You are already assigned to work from ${formatTime(
+          startTime
+        )} to ${formatTime(endTime)} on ${formatDate(startTime)}`;
+      } else {
+        console.log('here');
+        text = `You are already assigned a scheduled rest day on ${formatDate(
+          startTime
+        )}`;
+      }
+      return <p>{text}</p>;
+    }
+  };
+
+  const shiftClashToText = (clash) => {
+    var startTime = Date.parse(clash.startTime);
+    var endTime = Date.parse(clash.endTime);
+    var clashText = `${formatTime(startTime)} to ${formatTime(
+      endTime
+    )} on ${formatDate(startTime)}`;
+    return clashText;
   };
 
   const handleServerValidationErrors = (error) => {
@@ -56,13 +98,14 @@ const EditShiftHours = ({
     let errorsHandled = true;
 
     var firstError = error[0];
+    setClashingTimes(firstError.data);
+    setClashingProperty(firstError.field);
 
     if (firstError.field == 'startAndEndTime') {
       summaryErrors['shift-start-time'] = {
         message:
           'Your start and finish times must not overlap with another time period',
       };
-      setWholeShiftClash(true);
     } else if (firstError.field == 'startTime') {
       summaryErrors['shift-start-time'] = {
         message: 'Your start time must not overlap with another period',
@@ -148,15 +191,48 @@ const EditShiftHours = ({
 
     const combinedErrors = { ...existingErrors };
 
-    if (wholeShiftClash) {
+    if (clashingProperty == 'startAndEndTime') {
+      combinedErrors['shift-start-time'] = {
+        message: (
+          <div>
+            <p>
+              Your start and finish times must not overlap with another time
+              period
+            </p>
+            {formatTimeClashes(clashingTimes)}
+          </div>
+        ),
+      };
       combinedErrors['shift-finish-time'] = {
         message: '',
       };
     }
 
+    if (clashingProperty == 'startTime') {
+      combinedErrors['shift-start-time'] = {
+        message: (
+          <div>
+            <p>Your start time must not overlap with another period</p>
+            {formatTimeClashes(clashingTimes)}
+          </div>
+        ),
+      };
+    }
+    console.log(clashingProperty);
+    if (clashingProperty == 'endTime') {
+      combinedErrors['shift-finish-time'] = {
+        message: (
+          <div>
+            <p>Your end time must not overlap with another period</p>
+            {formatTimeClashes(clashingTimes)}
+          </div>
+        ),
+      };
+    }
+
     console.log('here');
     console.log(combinedErrors);
-    return combinedErrors; //Object.keys(errors).length > 0 ? errors : summaryErrors;
+    return combinedErrors;
   };
 
   return (
--- a/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
+++ b/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
@@ -143,15 +143,20 @@ const EditShiftHours = ({
   };
 
   const getCombinedErrors = () => {
-    const combinedErrors = deepCloneJson(
-      Object.keys(errors).length > 0 ? errors : summaryErrors
-    );
+    var existingErrors =
+      Object.keys(errors).length > 0 ? errors : summaryErrors;
+
+    const combinedErrors = { ...existingErrors };
+
     if (wholeShiftClash) {
       combinedErrors['shift-finish-time'] = {
         message: '',
       };
     }
-    return combinedErrors;
+
+    console.log('here');
+    console.log(combinedErrors);
+    return combinedErrors; //Object.keys(errors).length > 0 ? errors : summaryErrors;
   };
 
   return (
--- a/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
+++ b/src/components/timecard/edit-shift-hours/EditShiftHours.jsx
@@ -87,7 +87,7 @@ const EditShiftHours = ({
         setClashingTimes(error.data);
         setClashingProperty(error.field);
         summaryErrors['shift-finish-time'] = {
-          message: 'Your end time must not overlap with another time period',
+          message: 'Your finish time must not overlap with another time period',
         };
       } else {
         errorsHandled = false;
@@ -171,39 +171,23 @@ const EditShiftHours = ({
 
     const combinedErrors = { ...existingErrors };
 
-    if (clashingProperty === 'startAndEndTime') {
+    var timeInputErrors = (
+      <TimeInputErrors
+        clashingProperty={clashingProperty}
+        clashes={clashingTimes}
+      />
+    );
+
+    if (clashingProperty === 'startTime' || clashingProperty === 'endTime') {
       combinedErrors['shift-start-time'] = {
-        message: (
-          <TimeInputErrors
-            clashingProperty={clashingProperty}
-            clashes={clashingTimes}
-          />
-        ),
+        message: timeInputErrors,
+      };
+    } else if (clashingProperty === 'startAndEndTime') {
+      combinedErrors['shift-start-time'] = {
+        message: timeInputErrors,
       };
       combinedErrors['shift-finish-time'] = {
         message: '',
-      };
-    }
-
-    if (clashingProperty === 'startTime') {
-      combinedErrors['shift-start-time'] = {
-        message: (
-          <TimeInputErrors
-            clashingProperty={clashingProperty}
-            clashes={clashingTimes}
-          />
-        ),
-      };
-    }
-
-    if (clashingProperty === 'endTime') {
-      combinedErrors['shift-finish-time'] = {
-        message: (
-          <TimeInputErrors
-            clashingProperty={clashingProperty}
-            clashes={clashingTimes}
-          />
-        ),
       };
     }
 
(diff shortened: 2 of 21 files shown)


LaurieBennett (changes requested):

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +160,62 @@ const EditShiftHours = ({
...
     );
   };
 
+  const getCombinedErrors = () => {
+    const existingErrors =
+      Object.keys(errors).length > 0 ? errors : summaryErrors;
+
+    const combinedErrors = { ...existingErrors };
+
+    if (clashingProperty == 'startAndEndTime') {
```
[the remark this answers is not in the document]
claire-guest: I've changed all of these now 👍

LaurieBennett (dismissed):

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -59,20 +62,37 @@ const EditShiftHours = ({
...
+        setClashingProperty(error.field);
+        summaryErrors['shift-start-time'] = {
+          message: 'Your start time must not overlap with another time period',
+        };
+      } else if (error.field === 'endTime') {
+        setClashingTimes(error.data);
+        setClashingProperty(error.field);
+        summaryErrors['shift-finish-time'] = {
+          message: 'Your end time must not overlap with another time period',
+        };
```
daleelR, line 91: Is it possible to overlap with multiple finish times?
E.g. There exists 2 complete time entries on a Wednesday. I save a new time entry on e.g. Tuesday 11pm to Thursday 1am.

Technically i'd clash with those 2 finish times on Wednesday right?
If that is a valid scenario to display all those overlapping finish periods in the error message I think this would only show the last one.

But maybe that scenario isn't valid/relevant in which case it's fine. 
I guess my point is, are we sure we only have to ever account for a single clashing finish time?

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -59,20 +62,37 @@ const EditShiftHours = ({
...
+      } else {
+        errorsHandled = false;
```
daleelR, line 93: We only hit this if the server has given us some dodgy data e.g a clashing entry without an expected error.field right..
If so, I think that means we can break the loop here as we're just going to show the service error.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -59,20 +62,37 @@ const EditShiftHours = ({
     }
   };
 
-  const handleServerValidationErrors = (error) => {
+  const handleServerValidationErrors = (errors) => {
+    if (errors === null || !Array.isArray(errors)) {
```
daleelR, line 66: Why do we need the array check here?
JpaRest always returns an array for the validation errors (empty if no errors) doesn't it? 
Wouldn't the null check suffice?

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +160,62 @@ const EditShiftHours = ({
...
     );
   };
 
+  const getCombinedErrors = () => {
+    const existingErrors =
+      Object.keys(errors).length > 0 ? errors : summaryErrors;
+
+    const combinedErrors = { ...existingErrors };
+
+    if (clashingProperty === 'startAndEndTime') {
```
daleelR, line 174: Can we extract this string and the other two below into somewhere common and then use those constants here and also in the `handleServerValidationErrors` function.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +160,62 @@ const EditShiftHours = ({
...
+      combinedErrors['shift-start-time'] = {
+        message: (
+          <TimeInputErrors
+            clashingProperty={clashingProperty}
+            clashes={clashingTimes}
+          />
+        ),
+      };
```
daleelR, line 182: We could compress the code down a bit as this object is repeated 3 times, we could assign in to a variable before hand and then just assign it to these keys.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +160,62 @@ const EditShiftHours = ({
...
+      combinedErrors['shift-finish-time'] = {
+        message: '',
+      };
+    }
+
+    if (clashingProperty === 'startTime') {
```
daleelR, line 188: Shouldn't these be `else if` statements as the `clashingProperty` can only be one of those 3 values at a time..

src/components/timecard/time-input-errors/TimeInputErrors.jsx
```diff
@@ -0,0 +1,118 @@
...
+      clash.startTime
+    )}`;
+  };
+
+  const startDateTime = (clash) => {
+    const startTime = new Date(clash.startTime);
+    const endTime = new Date(clash.endTime);
+    let startDate = '';
+
+    if (startTime.toDateString() !== endTime.toDateString()) {
```
daleelR, line 91: any specific reason for converting the date to a textual string and then comparing?
Can we use `getTime()` instead?

src/components/timecard/time-input-errors/TimeInputErrors.jsx
```diff
@@ -0,0 +1,118 @@
...
+        clash.endTime
+      )} on ${formatLongDate(clash.endTime)}`;
+    }
+
+    return `${timePeriodType.toLowerCase()} on ${formatLongDate(
+      clash.startTime
+    )}`;
+  };
+
+  const startDateTime = (clash) => {
```
daleelR, line 86: nit: This name suggest an actual dateTime value. It actually returns a formatted textual message inc the start date and time.
Can we improve the name slightly

src/components/timecard/time-input-errors/TimeInputErrors.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+    return (
+      <p>
+        Your start and finish times must not overlap with another time period
+      </p>
+    );
+  };
+
+  const displaySingleTimeClash = () => {
+    const clash = clashes[0];
```
daleelR, line 30: nit: optional, could throw an error with a nice message before this if there are no clashes. In the rare circumstance that this component gets called an empty clashes array.

src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
```diff
@@ -7,7 +7,6 @@ const NotesSummaryList = () => {
   const [addNote, setAddNote] = useState(false);
 
   const handleClick = () => {
-    event.preventDefault();
```
daleelR, line 10: why don't we need this anymore?
Doesn't the default link event behavior cause a refresh or something like that?

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: I'm not sure I follow? I've made a scenario where the finish time clashes with two other shift and it seems to work as intended, unless this isn't the situation you mean?
Screenshot 2022-11-14 at 10 00 19

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: For any other kind of error from the server that is not validation, the data will have the format `message: "some string"`, so if we don't check for an array here, it will throw an error when it gets to the forEach on line 72.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: 👍  done

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: 👍  done

src/components/timecard/time-input-errors/TimeInputErrors.jsx
claire-guest: This is checking if they are on the same day, rather than if they are the same day and time. We decided against using `getDate` as it only returns the day of the month, not an actual date. We could do something like this (https://stackoverflow.com/a/43855248), but it seemed quite long winded?

src/components/timecard/time-input-errors/TimeInputErrors.jsx
claire-guest: 👍  done

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
daleelR: Great, yeah I think that's the scenario I was curious about. Looks like it's all handled. Will resolve.

src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
claire-guest: I know sonarcloud was highlighting it as a code smell, and Joanna thought we didn't need it, but I'm not sure of the exact reason. I can put it back in as it doesn't really affect this PR, and can be looked into at a later date?

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
daleelR: Ah I see, yep that makes sense.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: I've made this change, do you think this function is harder to read now, with just the boolean returns and no variable name to indicate what they mean, or is it still ok to follow?

src/components/timecard/time-input-errors/TimeInputErrors.jsx
daleelR: Right I see, can we possibly use dayJs instead and use the `isSame` function and pass in `day` as the second param. 
It looks like that would internally compare day, month, and year (like SO ex) but in a concise line for us. Plus it would keep consistency & simplicity with using dayJs library.
https://day.js.org/docs/en/query/is-same

If that somehow ends up consequently over complicating other parts then I think it's fine as it is. But yeah might be worth a try.

src/components/common/form/text-area/notes-summary-list/NotesSummaryList.jsx
daleelR: oh right, looks like you guys confirmed it's all good so no worries. Will resolve.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
daleelR: I don't think it's become particularly more difficult to follow. 
Buttt if you wanted to maintain that extra level of readability, you could use a `for of `loop instead and then `break` instead of return false.
E.g. (untested)
```
for (const error of errors) {
      if (error.field === 'startAndEndTime') {
        setClashData(
          'shift-start-time',
          'Your start and finish times must not overlap with another time period'
        );
      } else if (error.field === 'startTime') {
        setClashData(
          'shift-start-time',
          'Your start time must not overlap with another time period'
        );
      } else if (error.field === 'endTime') {
        setClashData(
          'shift-finish-time',
          'Your finish time must not overlap with another time period'
        );
      } else {
        errorsHandled = false;
        break;
      }
    }
```

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +156,54 @@ const EditShiftHours = ({
           setShowEditShiftHours(false);
         }
       },
+      true,
       handleServerValidationErrors
     );
   };
 
+  const getCombinedErrors = () => {
```
holly-blues, line 164: Consider lifting this function out of here with its own tests/test file. EditShiftHours seems to be doing a lot and this concern could be separated out.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +156,54 @@ const EditShiftHours = ({
...
+    const existingErrors =
+      Object.keys(errors).length > 0 ? errors : summaryErrors;
+
+    const combinedErrors = { ...existingErrors };
+
+    var timeInputErrors = (
```
holly-blues, line 170: Prefer `let` here

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -140,16 +156,54 @@ const EditShiftHours = ({
...
+        clashingProperty={clashingProperty}
+        clashes={clashingTimes}
+      />
+    );
+
+    if (clashingProperty === 'startAndEndTime') {
+      combinedErrors['shift-start-time'] = {
+        message: timeInputErrors,
+      };
+      combinedErrors['shift-finish-time'] = {
```
holly-blues, line 181: Per other constants comment, consider putting these control names in a constants file. They are now getting used in many places
`shift-start-time`, `shift-finish-time`

src/components/timecard/time-input-errors/TimeInputErrors.jsx
```diff
@@ -0,0 +1,123 @@
...
+              {timePeriodClashToText(clash)}
+            </li>
+          ))}
+        </ul>
+      </div>
+    );
+    return times;
+  };
+
+  const sortClashes = (clashes) => {
```
holly-blues, line 71: Consider renaming this (e.g. `sortByStartTime`) to let future code maintainers understand its purpose, and moving to a utility if it could be reused elsewhere in the application

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: Unresolving this as I've just changed it back; moving the object outside was raising an error because it was trying to create the object with no data when there were no time clash errors, and I can't think of a way to avoid it 🤔

src/components/timecard/time-input-errors/TimeInputErrors.jsx
claire-guest: I've renamed it, I think it might be too specific to be reused elsewhere?

src/components/timecard/time-input-errors/TimeInputErrors.jsx
claire-guest: Updated to use dayjs :)

src/components/timecard/time-input-errors/TimeInputErrors.jsx
holly-blues: Looks good. Ok - just a thought :)

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: 👍 done

src/components/timecard/time-input-errors/TimeInputErrors.jsx
claire-guest: 👍 done

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: 👍 done

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: I like this, have made the change now :)

src/components/timecard/time-input-errors/TimeInputErrors.jsx
daleelR: Nice, minor further suggestion is to remove the javascript date objects too (ln 94,95). 
I think you should be able to just pass in the times from the clash object to dayjs directly.
E.g
`if (!dayjs(clash.startTime).isSame(dayjs(clash.endTime), 'day')) {`

Again feel free to just resolve if short for time / doesn't work.

src/components/timecard/time-input-errors/TimeInputErrors.jsx
daleelR: nice. just a hint (no need to change), `==` would have been sufficient on ln 32 as array.length always returns an integer.

src/utils/constants.js
```diff
@@ -0,0 +1,10 @@
+export const clashingProperties = {
```
daleelR, line 1: Slight issue with the way these are defined. The object has been declared as a constant but the values are still mutable and can be overwritten by anywhere in the code effectively making them not constants.
Might be better to define them individually e.g
```
const START_TIME = 'startTime';
const END_TIME = 'endTime';
etc.
```
Also allows us to then import only the ones which we need.

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
claire-guest: 👍 done this now, though now sure where it should live? I've put it in utils for now, interested in your thoughts!

src/utils/constants.js
claire-guest: I see what you mean, though it would still be nice to group them in some way. Do you think using Object.freeze would be an ok alternative? Pro: it allows grouping, con: you still have to import a whole group, not an individual constant?

src/components/timecard/time-input-errors/TimeInputErrors.jsx
claire-guest: 👍 done

src/utils/constants.js
claire-guest: updated :)

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
holly-blues: Looks great and I like it and where you have put it

holly-blues (approved):

src/utils/time-entry-utils/combineTimeErrors.spec.jsx
```diff
@@ -0,0 +1,86 @@
+import TimeInputErrors from '../../components/timecard/time-input-errors/TimeInputErrors';
+import { clashingProperties, inputNames } from '../constants';
+import { combineExistingAndTimeClashErrors } from './combineTimeErrors';
+
+describe('combineExistingAndTimeClashErrors', () => {
+  it('should return an empty array when no errors are passed in', async () => {
+    var result = combineExistingAndTimeClashErrors({}, {}, null, null);
```
laurieboyes, line 7: Quibble: Is there any reason to use `var` instead of `const` in these bits? I think the difference in functionality are but `const` is more idiomatic

laurieboyes: Agreed we’ll fix this later for the sake of not having to re-request reviews 😄

PR #48, opened 2022-11-11
Eahw 589 fix the api filter functionality

JWeston-HO:
This PR is to address the filter functionality required for eahw-589.
The requirement is to show time entries that cover two days to show on the day they start and on the day they end.

### DIFF
--- a/src/utils/api-utils/UrlSearchParamBuilder.js
+++ b/src/utils/api-utils/UrlSearchParamBuilder.js
@@ -10,8 +10,8 @@ export class UrlSearchParamBuilder {
     return this;
   }
 
-  setFilters(...filters) {
-    this.params.append('filter', filters.join('&&'));
+  setFilters(filter) {
+    this.params.append('filter', filter);
     return this;
   }
 
@@ -19,3 +19,13 @@ export class UrlSearchParamBuilder {
     return this.params;
   }
 }
+
+export const joinWithAnd = (...filters) => {
+  console.log(filters.join('&&'));
+  return filters.join('&&');
+};
+
+export const joinWithOr = (...filters) => {
+  console.log(filters.join('||'));
+  return filters.join('||');
+};
--- a/src/utils/api-utils/UrlSearchParamBuilder.js
+++ b/src/utils/api-utils/UrlSearchParamBuilder.js
@@ -10,7 +10,7 @@ export class UrlSearchParamBuilder {
     return this;
   }
 
-  setFilters(filter) {
+  setFilter(filter) {
     this.params.append('filter', filter);
     return this;
   }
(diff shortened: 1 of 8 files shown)


src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
```diff
@@ -2,13 +2,36 @@ import dayjs from 'dayjs';
 import {
   filterOnOrAfterDate,
   filterBeforeDate,
+  joinOrConditions,
+  joinAndConditions,
 } from '../resource-filter-builder/resourceFilterBuilder';
 
-export const filterTimeEntriesOnDate = (date) => {
-  const startOfDateFilter = filterOnOrAfterDate('actualStartTime', date);
+export const filterTimeEntriesOnDate = (resourceDateProperty, date) => {
```
daleelR, line 9: we no longer need to export this function as `buildTimeCardFilter` is the one used by `TimeCard` now

src/utils/api-utils/UrlSearchParamBuilder.js
```diff
@@ -10,8 +10,8 @@ export class UrlSearchParamBuilder {
     return this;
   }
 
-  setFilters(...filters) {
-    this.params.append('filter', filters.join('&&'));
+  setFilters(filter) {
```
daleelR, line 13: nit: could this be named `setFilter` without the 's'? Then if we ever need to re-introduce the above removed function for multiple separate filters we can with it's appropriate name.

src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
```diff
@@ -2,13 +2,36 @@ import dayjs from 'dayjs';
...
+  const startOfDateFilter = filterOnOrAfterDate(resourceDateProperty, date);
   const endOfDateFilter = filterBeforeDate(
-    'actualStartTime',
+    resourceDateProperty,
     dayjs(date).add(1, 'day')
   );
   return [startOfDateFilter, endOfDateFilter];
 };
+
+export const buildTimeCardFilter = (date, userId) => {
```
daleelR, line 18: nit: should probs be named buildTime**Entries**Filter instead.

src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
JWeston-HO: export removed.

src/utils/api-utils/UrlSearchParamBuilder.js
JWeston-HO: Changed to setFilter

daleelR (approved):

LaurieBennett (approved):

PR #49, opened 2022-11-14
Switch to notprod domain APIs

laurieboyes:
This _should_ switch over without downtime

Our APIs are currently working on the new notprod domain: https://timecard.dev.callisto-notprod.homeoffice.gov.uk/swagger-ui/index.html

And our ingress is configured so that each API service enables CORS for both of our UI domains:
[Timecard API cors settings](https://github.com/UKHomeOffice/callisto-ingress-nginx/blob/main/helm/templates/ingress_timecard.yml#L15)
[What those values are set to](https://github.com/UKHomeOffice/callisto-ingress-nginx/blob/main/helm/values.yaml#L6-L7)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -73,8 +73,8 @@ steps:
         from_secret: kc_realm_dev
       VITE_KC_CLIENTID:
         from_secret: kc_clientid_dev
-      VITE_TIMECARD_API_URL: "https://timecard.dev.callisto.homeoffice.gov.uk/"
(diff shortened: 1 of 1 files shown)


JWeston-HO (approved):

daleelR (approved):

laurieboyes:
Closed because I used an emoji in the description and it broke the drone integration 🙄

PR #54, opened 2022-11-22
add owner id to end time filter

JWeston-HO:
Add owner id to the end time filter. Time periods could be seen by other users. By adding the owner id to the end time filter

### DIFF
--- a/src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
+++ b/src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
@@ -16,7 +16,7 @@ const filterTimeEntriesOnDate = (resourceDateProperty, date) => {
 };
 
 export const buildTimeEntriesFilter = (date, userId) => {
-  const ownerAndUserId = "ownerId=='" + userId + "'";
+  const ownerAndUserId = `ownerId=='${userId}'`;
 
   const startDateFilterCondition = joinAndConditions(
     ownerAndUserId,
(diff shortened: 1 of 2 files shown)


src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
```diff
@@ -17,10 +17,12 @@ const filterTimeEntriesOnDate = (resourceDateProperty, date) => {
 
 export const buildTimeEntriesFilter = (date, userId) => {
   const startDateFilterCondition = joinAndConditions(
+    "ownerId=='" + userId + "'",
     ...filterTimeEntriesOnDate('actualStartTime', date)
```
laurieboyes, line 21: Not sure if the line breaks will go weird but I’m a fan of template strings over concatenation. Would this work?
```suggestion
    `ownerId=='${userId}'${...filterTimeEntriesOnDate('actualStartTime', date)}`
```

JWeston-HO: Not sure the above would work. We need to add the && in-between filter arguments. I have pushed a changed that stores "ownerId=='" + userId + "'" as a const. As this is used multiple times.

src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
```diff
@@ -16,11 +16,15 @@ const filterTimeEntriesOnDate = (resourceDateProperty, date) => {
 };
 
 export const buildTimeEntriesFilter = (date, userId) => {
+  const ownerAndUserId = "ownerId=='" + userId + "'";
```
laurieboyes, line 19: It’s an even more pedantic comment than it was before but could _this_ be a template string?
```suggestion
  const ownerAndUserId = `ownerId==${userId}'`;
```

JWeston-HO: yes it can committed.

laurieboyes (approved):

joannafawl (approved):

PR #56, opened 2022-11-23
Eahw 2337/promote prod

holly-blues:
These are changes to the UI build pipelines to enable a manual promotion to Test. Manual promotion to Production will be a later PR depending on how this goes. Summary of the main changes are:

- Exclude the Build pipeline (clone, fetch, install, lint, test, sonar & build) from triggering on Promote and Tag events
- Create a new Dev deploy pipeline with a new step to git tag the build using the build number (for now, semver later), then do the upload to S3 step. This new pipeline is only triggered on push to the Main branch e.g. pull request merge. It has a dependency on the Build pipeline succeeding
- Create a new Test deploy pipeline with new steps to: (1) build for Test using the Test environment variables (2) upload to the dev S3 bucket Test folder. This depends on the previous Dev deploy pipeline success.
- Add a new notification for the new Promote step which is only triggered by the Promote event

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -98,10 +98,11 @@ steps:
     when:
       event: 
         - push
-      # branch:
-      #   - main
+      branch:
+        - main
     depends_on:
       - build project
+      - sonar analysis      
   
   - name: upload to S3
     image: plugins/s3
--- a/.drone.yml
+++ b/.drone.yml
@@ -212,7 +212,7 @@ steps:
     settings:
       template: |
         {{#success build.status}}
-          Build #{{ build.number }} succeeded! :tada: {{#main build.branch}}and deployed to Dev{{/main}}
+          Build #{{ build.number }} succeeded {{#main build.branch}}and deployed to Dev{{/main}}! :tada:
         {{else}}
           Build #{{ build.number }} failed :alert:
         {{/success}}
--- a/.drone.yml
+++ b/.drone.yml
@@ -271,7 +271,7 @@ steps:
         {{#success build.status}}
           :rocket: Successful *{{uppercasefirst build.deployTo}}* deployment for *{{repo.name}} build #{{build.number}}*.          
         {{else}}
-          :zombie: Problem *{{uppercasefirst build.deployTo}}* deployment failed for <${DRONE_BUILD_LINK}| build #{{build.number}}>.
+          :zombie: Problem *{{uppercasefirst build.deployTo}}* deployment failed for *{{repo.name}}* <${DRONE_BUILD_LINK}| build #{{build.number}}>.
         {{/success}}       
     when:
       event: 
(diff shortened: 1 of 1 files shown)


taomoh (commented):
The back-end projects use a different approach for build and deployment:

- The project is built into a Docker image which is pushed to ECR:
https://github.com/UKHomeOffice/callisto-timecard-restapi/blob/5941dc6481765051cceb2816e1032c418ce7d6d0/.drone.yml#L59

- The deployment is using the image from ECR created in step above:
https://github.com/UKHomeOffice/callisto-timecard-restapi/blob/5941dc6481765051cceb2816e1032c418ce7d6d0/.drone.yml#L149

I think we should aim to homogenise the process across projects. I appreciate this wasn't the purpose of this PR so we can tackle this separately

joannafawl (approved):

.drone.yml
```diff
@@ -92,44 +128,112 @@ steps:
...
+platform:
+  os: linux
+  arch: amd64
+
+depends_on:
+  - callisto-ui-dev-deploy
+
+trigger:
+  event:
+    - promote
```
daleelR, line 148: Does this alone prevent us from being able to promote a build from any branch that isn't main?
Do we need something else to only allow promoting once it's pushed to main, or was that to come separately?

holly-blues:
> The back-end projects use a different approach for build and deployment:
> 
> * The project is built into a Docker image which is pushed to ECR:
>   https://github.com/UKHomeOffice/callisto-timecard-restapi/blob/5941dc6481765051cceb2816e1032c418ce7d6d0/.drone.yml#L59
> * The deployment is using the image from ECR created in step above:
>   https://github.com/UKHomeOffice/callisto-timecard-restapi/blob/5941dc6481765051cceb2816e1032c418ce7d6d0/.drone.yml#L149
> 
> I think we should aim to homogenise the process across projects. I appreciate this wasn't the purpose of this PR so we can tackle this separately

@taomoh 
I definitely agree to a certain extent, since containers are the ACP way and the path of least resistance.

However, the website deployment (documented in Confluence Development > Architecture > Deployment Diagram) is using a different approach.

It has been a troublesome approach and as you point out, an inconsistent one. It certainly won’t get addressed as part of this PR, but might be worth a discussion on the next Tech Huddle, especially given the new local development work is container-tastic :)

.drone.yml
holly-blues: Good shout. I was going to look into ways to protect only promoting from the main branch next, but I think the options might be limited with Drone...

daleelR: Ah ok, I just tested on the timecard one, if you include an event here for push for 'main' branch and then attempt to promote a from a non main branch although it will still run, it will not run the test depl pipeline.
So that's a partial solution in that it prevents the main stuff but unfortunately still allows us to trigger the promotion.

I'll resolve now anyway, perhaps we can discuss afterwards..

PR #58, opened 2022-11-24
EAHW-2429: Move tag and upload steps back to build pipeline

holly-blues:
Moving the tag and upload steps back into the build pipeline - to fix the issue where there are no files available to the pipeline to be uploaded. See screenshot below.

Also git tag is not pulling the drone build version through - trying tweaking the syntax.

calamity-drone

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -101,8 +101,7 @@ steps:
        - push
       branch:
        - main  
-    depends_on:
-      - install dependencies
+    depends_on:      
       - sonar analysis
   
   - name: upload to S3
--- a/.drone.yml
+++ b/.drone.yml
@@ -86,25 +86,6 @@ steps:
     depends_on:
       - install dependencies
 
----
-kind: pipeline
-type: kubernetes
-name: callisto-ui-dev-deploy
-
-platform:
-  os: linux
-  arch: amd64
-
-depends_on:
-  - callisto-ui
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -141,7 +135,7 @@ platform:
   arch: amd64
 
 depends_on:
-  - callisto-ui-dev-deploy
+  - callisto-ui
```
laurieboyes, line 138: Does this mean that the dev and test deploys will now happen in parallel?

laurieboyes: Oh wait no I get it 👍

laurieboyes (approved):

daleelR (dismissed):

.drone.yml
```diff
@@ -113,8 +94,16 @@ steps:
...
-      followtags: true     
+      tag: v${DRONE_BUILD_NUMBER}
+      followtags: true
+    when:
+      event:
+       - push
+      branch:
+       - main  
+    depends_on:
+      - install dependencies
```
daleelR, line 105: I don't think you need to depend on 'install dependencies' as 'sonar analysis' depends on 'test project' which in turn depends on 'install dependencies'.

daleelR (approved):

laurieboyes (approved):

PR #59, opened 2022-11-25
Eahw 2429/promote UI test

holly-blues:
- Add 'promote to test check' to ensure build being promoted is on the main branch
- Remove commit message on git tag
- [Example branch promotion failing the check](https://drone-gh.acp.homeoffice.gov.uk/UKHomeOffice/callisto-ui/843)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -93,7 +93,6 @@ steps:
       remote: git@github.com:UKHomeOffice/callisto-ui.git
       ssh_key:
         from_secret: github_private_key
-      commit: true
       tag: v${DRONE_BUILD_NUMBER}
       followtags: true
     when:
--- a/.drone.yml
+++ b/.drone.yml
@@ -101,8 +101,7 @@ steps:
        - push
       branch:
        - main  
-    depends_on:
-      - install dependencies
+    depends_on:      
       - sonar analysis
   
   - name: upload to S3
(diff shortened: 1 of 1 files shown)


joannafawl (approved):

.drone.yml
```diff
@@ -159,6 +168,8 @@ steps:
     commands:
       - npm ci
       - npm run build
+    depends_on:
+      - promote to test check
```
daleelR, line 172: Were you able to verify that this step runs when we are promoting on main?
I'm curious because with the timecard one when the branch is actually main the 'check' step doesn't actually run at all.
E.g. https://drone-gh.acp.homeoffice.gov.uk/UKHomeOffice/callisto-timecard-restapi/563

And cos you're using a `depends_on` i'm wondering if this will even run if the 'check' step doesn't actually run when the promotion is valid.

Might be worth verifying if it works if you haven't already.

holly-blues: Ah good idea, will test that out 👍

holly-blues: From my extensive local testing (ahem) it doesn't seem to matter if the step isn't run. Which is what the [docs](https://docs.drone.io/pipeline/ssh/syntax/parallelism/) suggest too: `"Note that you can use conditional steps in your dependency graph. The scheduler automatically corrects the dependency graph for skipped steps." `

So I think we're good 🤞

daleelR: ah cool, looks good then 👍

daleelR (approved):

PR #66, opened 2022-12-21
LocalDev - UI solution

sewerynmi:
Removed temp and not required files

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,3 +130,32 @@ VITE_ACCRUALS_API_URL_PROXY="http://localhost:9091/"
 ## Troubleshooting
 
 If you’re using an AMD (M1,M2) chip mac, you might encounter problems starting the keycloak docker container. We’ve had some success building a new image locally, as per the instructions in this github thread: https://github.com/docker/for-mac/issues/5310#issuecomment-877653653
+
+## Running UI in LocalDev Environment
+
+### Step 1
+
+Download Callisto LocalDev repository from https://github.com/UKHomeOffice/callisto-localdev
+
+Run its services with command : `docker compose up -d`
+
+### Step 2
+
+Stop (manually) service `web` (most likely the name will be `web-1`)
+
+### Step 3
+
+Create file .env in the root of this project, and add the following:
+
+```
+VITE_KC_URL="https://keycloak.callisto.localhost/auth/"
+VITE_KC_REALM="callisto"
+VITE_KC_CLIENTID="callistoreactclient"
+
+VITE_TIMECARD_API_URL = "https://timecard-restapi.callisto.localhost/"
+VITE_ACCRUALS_API_URL = "https://accruals-restapi.callisto.localhost/"
+```
+
+### Step 4
+
+Run command `docker compose up -d`
--- a/README.md
+++ b/README.md
@@ -135,27 +135,40 @@ If you’re using an AMD (M1,M2) chip mac, you might encounter problems starting
 
 ### Step 1
 
-Download Callisto LocalDev repository from https://github.com/UKHomeOffice/callisto-localdev
-
-Run its services with command : `docker compose up -d`
+Download Callisto LocalDev repository from https://github.com/UKHomeOffice/callisto-localdev and follow the instruction in the callisto-localdev repository to start local environment.
 
 ### Step 2
 
-Stop (manually) service `web` (most likely the name will be `web-1`)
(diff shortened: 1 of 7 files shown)


stephenkaye-ho (approved):
Looks. good

README.md
```diff
@@ -130,3 +130,32 @@ VITE_ACCRUALS_API_URL_PROXY="http://localhost:9091/"
...
 
 If you’re using an AMD (M1,M2) chip mac, you might encounter problems starting the keycloak docker container. We’ve had some success building a new image locally, as per the instructions in this github thread: https://github.com/docker/for-mac/issues/5310#issuecomment-877653653
+
+## Running UI in LocalDev Environment
+
+### Step 1
+
+Download Callisto LocalDev repository from https://github.com/UKHomeOffice/callisto-localdev
+
+Run its services with command : `docker compose up -d`
```
stephenkaye-ho, line 140: I think this should be changed to tell them to follow the instructions in the callisto-localdev readme

README.md
```diff
@@ -130,3 +130,32 @@ VITE_ACCRUALS_API_URL_PROXY="http://localhost:9091/"
...
+
+### Step 2
+
+Stop (manually) service `web` (most likely the name will be `web-1`)
```
stephenkaye-ho, line 144: We should change this to use the docker compose method
`docker compose stop web`

README.md
```diff
@@ -130,3 +130,32 @@ VITE_ACCRUALS_API_URL_PROXY="http://localhost:9091/"
...
+
+### Step 3
+
+Create file .env in the root of this project, and add the following:
```
stephenkaye-ho, line 148: If this is needed then we should check in the .env file with these settings and remove it from the git ignore. The aim here is zero config. If localdev is in use, these values wont change. If we are concerned about conflicting with current development not using docker compose (localdev) then we should use a different file name e.g. docker.env and then use the [env_file configuration option](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option)

stephenkaye-ho (changes requested):

README.md
sewerynmi: sure, let's use docker compose for that
Fixed.

README.md
sewerynmi: I guess I will change it to use 2 files:  .env and .env.local
We're going to checkout .env but not .env.local which will be specific for each developer. 
LocalDev variables will be stored in the .env file. This one shouldn't be edited by developers. 
Variables for a solution without using LocalDev will be stored in the .env.local which takes precedence over the .env file. Developers can do anything they want in that file. For example: enabling the usage of mock data etc. 
Is that ok?

stephenkaye-ho: So this bit won't be needed, we'll check in the .env file with correct settings and remove it from the .gitignore file. I don't know if we need to document the .env.local here or if that's just a common thing to document elsewhere. I presume it's taking advantaged of docker using .env whilst the react app falls back from .env.local to .env allowing us to use it as an override/alternative

package.json
```diff
@@ -4,17 +4,19 @@
...
+    "start:jsonserver": "node mocks/apiServer.js",
     "test": "TZ=UTC jest --coverage",
     "monitor": "TZ=UTC jest --watchAll",
     "lint:fix": "eslint --ext .js,.jsx src --fix",
     "lint": "eslint --ext .js,.jsx src",
     "build": "vite build",
     "preview": "vite preview",
-    "vite": "vite"
+    "vite": "vite",
+    "localdev": "run-p localdev:init start:jsonserver vite ",
```
stephenkaye-ho, line 18: should this be `run-p localdev jsonserver vite`
and we should be using pre/post prefixes for the init. Should we always start the jsonserver (that's probably a sensible choice). I think we should tidy up these scripts and drop out keycloak. The timezone env var is probably a good candidate for also going in the .env file

package.json
```diff
@@ -4,17 +4,19 @@
...
+    "localdev:init": "npm install"
```
stephenkaye-ho, line 19: When is this used?

stephenkaye-ho (changes requested):

package.json
stephenkaye-ho, line 18: this also fails for me on a clean build because 
```
sh: run-p: not found
```

stephenkaye-ho: Realised some other issue with the npm script already depending on run-p before npm install is guaranteed to have been called so moved this into the entry point as a shell command. This means there are no changes to the script section so the configuration/structure can be assessed elsewhere as this PR no longer touches them

stephenkaye-ho (approved):

PR #72, opened 2023-01-23
Eahw 2554/UI branch deploys

joannafawl:
- Deploys branch by pushing to S3 and deploying ingress for branch on pull request
- Branch deployments can be accessed on `eahw-xxxx-web.dev.callisto-notprod.homeoffice.gov.uk`, eg. [here](https://eahw-2554-web.dev.callisto-notprod.homeoffice.gov.uk/)

Note: currently it just displays a white screen as it is trying to connect to Keycloak


.drone.yml
```diff
@@ -84,7 +84,26 @@ steps:
     commands:
       - npm run build
     depends_on:
-      - install dependencies
+      - lint project
+      - test project
+
+  - name: deploy branch to S3
```
holly-blues, line 90: Does it make sense to put this step into the new branch deploy pipeline? I think I just added this here for testing purposes. Also the depends_on changes above can be reverted.

joannafawl: I thought about doing that and tried putting the deploy to S3 step in different pipelines for branch and dev, but I think because the same files are used in branch and dev, we'd need to copy the install dependencies and build steps into two pipelines. I wasn't sure if that would be preferred to have repeated code if it means the pipeline is more clearly separated? 

For the depends_on, I thought we would only want the deploy to S3 to happen if tests pass, but maybe that logic could be moved

holly-blues: Ah yes. The node_modules won't be available to the new pipeline...

The proposed changes for Keycloak will require a branch specific Keycloak url, so will probably have to re-build for a branch but can cross that bridge then.

Happy to leave it for now.

holly-blues (approved):

JWeston-HO (approved):

PR #82, opened 2023-02-23
Eahw 1677 show time period type select when last period removed

LaurieBennett:


LaurieBennett:
Closed due to incorrect branch name

PR #86, opened 2023-03-10
Eahw590/finish time 24 hours after start date

joannafawl:


src/utils/filters/time-entry-filter/timeEntryFilterBuilder.js
```diff
@@ -7,10 +7,13 @@ import {
 } from '../resource-filter-builder/resourceFilterBuilder';
 
 const filterTimeEntriesOnDate = (resourceDateProperty, date) => {
```
daleelR, line 9: This function now builds filters specifically for 1 day before and 2 days after the specified date.
So we probs need to change this function name to represent that it's giving us entries on **& near** the date.
Or make the name represent a range and pass in the date boundaries.

src/pages/timecard/Timecard.jsx
```diff
@@ -154,22 +154,31 @@ const updateTimeEntryContextData = async (
...
-            timeEntry.actualEndTime ? timeEntry.actualEndTime : '',
-            timeEntry.timePeriodTypeId,
-            timeEntry.finishNextDay ??
-              isFinishTimeOnNextDay(
-                formatTime(timeEntry.actualStartTime),
-                formatTime(timeEntry.actualEndTime)
-              )
+          return !(
+            dayjs(timeEntry.actualEndTime).isBefore(timeCardStart) ||
+            dayjs(timeEntry.actualStartTime).isAfter(timeCardEnd)
```
daleelR, line 165: I know we spoke about this during the mob programming session, but now that I can see the finished filter
I **think** we may be able to include it all in the jpaRest request filter. Lets catchup and take another look to see if possible.
(Above comment becomes redundant if this can be done)

src/pages/timecard/Timecard.jsx
```diff
@@ -154,22 +154,31 @@ const updateTimeEntryContextData = async (
   validateServiceErrors(setServiceError, async () => {
     const timeEntriesResponse = await getTimeEntries(timeEntriesParams);
 
+    const timeCardStart = dayjs(date).startOf('day').add(1, 'minute');
+    const timeCardEnd = dayjs(date).endOf('day');
+
     if (timeEntriesResponse.data.items?.length > 0) {
-      const existingTimeEntries = timeEntriesResponse.data.items.map(
+      const filteredTimeEntries = timeEntriesResponse.data.items.filter(
```
daleelR, line 161: if we revisit the filter after a while it might not be easy to workout why we are doing this specific filtering, maybe we should update this variable name to indicate what's actually been filtered.
Maybe something that represents "time entries that span onto date".

daleelR: resolving as the additional filtering has now been removed.

src/pages/timecard/Timecard.jsx
daleelR: resolving as the additional filtering has now been removed.

LaurieBennett:
Closed to clean up commits due to opening this early

PR #88, opened 2023-03-22
Eahw2601/timecard changes

Xeena-Newman:


src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
```diff
@@ -68,9 +69,18 @@ const EditShiftTimecard = ({
...
           </dt>
-          <dd className="govuk-summary-list__value"></dd>
+          <dd
+            className="govuk-summary-list__value"
+            style={{ whiteSpace: 'nowrap' }}
+          >
+            {timeEntryExists
+              ? timecardDate === formatDate(timeEntry.startTime)
+                ? 'Shift'
+                : 'Shift (continued)'
```
daleelR, line 81: MIght be worth using the getting the 'Shift' text from the map returned by the `getTimePeriodTypesMap` function as that comes from the API. Easier maintenance.

IanFisherHO (changes requested):

src/components/timecard/edit-shift-timecard/EditShiftTimecard.spec.jsx
daleelR, line 1: Can we have tests to cover the new logic that shows "HH:MM **on DD Month** to" when the shift covers multiple days.
And also tests to cover the logic of conditionally showing the 'continued' text on shifts that cover multiple days.

src/components/timecard/timecard-entries-list/TimecardEntriesList.spec.jsx
```diff
@@ -22,7 +22,7 @@ describe('TimecardEntriesList', () => {
     expect(screen.queryByText('Add a new time period')).toBeFalsy();
     expect(screen.getByText('Start time')).toBeTruthy();
     expect(screen.getByText('Finish time')).toBeTruthy();
-    expect(screen.getByText('Shift')).toBeTruthy();
```
daleelR, line 25: Could you not still keep this since the test is specifically testing for when the period type is shift?

src/pages/timecard/Timecard.spec.jsx
```diff
@@ -80,6 +80,48 @@ describe('Shift spanning mutiple days', () => {
...
+        },
+      ],
+      setTimeEntries: jest.fn(),
+      timecardDate: '',
+      setTimecardDate: jest.fn(),
+    });
+
+    expect(
+      screen.getByText((content) =>
+        content.includes('01:00 on 1 September to 05:00 on 3 September')
```
daleelR, line 101: I think this can conflict with the test name. The test name suggests that it just needs to display the start and finish time when they have both been entered.
But this particular text we'd only want to show when covering multiple days. 
Maybe the test name should be updated.

src/pages/timecard/Timecard.spec.jsx
```diff
@@ -80,6 +80,48 @@ describe('Shift spanning mutiple days', () => {
...
 
+  it('should display start and finish time on timecard when both times have been entered', async () => {
+    renderWithTimecardContext(<Timecard />, {
+      summaryErrors: {},
+      setSummaryErrors: jest.fn(),
+      timeEntries: [
+        {
+          timePeriodTypeId: '00000000-0000-0000-0000-000000000001',
+          startTime: '2022-09-01 01:00:00+00:00',
+          finishTime: '2022-09-03 05:00:00+00:00',
```
daleelR, line 91: Assuming this test is actually specifically to cover those covering multiple days, it might be better to test the boundaries of the logic/requirement by having the start time be 11:59 and the finish time 00:00 the following day.

src/pages/timecard/Timecard.spec.jsx
```diff
@@ -80,6 +80,48 @@ describe('Shift spanning mutiple days', () => {
...
+    });
+
+    expect(
+      screen.getByText((content) =>
+        content.includes('01:00 on 1 September to 05:00 on 3 September')
+      )
+    ).toBeTruthy();
+  });
+
+  it('should display Shift(continued) if timecard spans more than one date and we are not on start date', async () => {
```
daleelR, line 106: Can we add a similar specific test to show that continued doesn't show when it spans more that one day but we're on the start date.

src/pages/timecard/Timecard.spec.jsx
Xeena-Newman: my mistake, that's been changed now

src/pages/timecard/Timecard.spec.jsx
Xeena-Newman: that's been changed now

src/pages/timecard/Timecard.spec.jsx
Xeena-Newman: That's been added now

src/components/timecard/timecard-entries-list/TimecardEntriesList.spec.jsx
daleelR: Ah ok, yes that makes sense, was just curious. Will resolve.

src/components/timecard/edit-shift-timecard/EditShiftTimecard.jsx
Xeena-Newman: a change has been made...it's caused 17 tests to fail, but is this change you meant?

Xeena-Newman: This will be revisited as a sub-task after the ui revamp

daleelR: Yeah that was the change, resolving after discussion and should be picked up as part of the revamp.

IanFisherHO (approved):

src/pages/timecard/Timecard.spec.jsx
daleelR: Might be a bit better to also expect that "(continued)" is falsy in the new test.

daleelR (approved):

PR #94, opened 2023-04-27
Eahw2742/shift clash no error

Xeena-Newman:


src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -212,12 +219,6 @@ const EditShiftHours = ({
...
             .setFinishNextDay(timeEntry.finishNextDay);
 
-          if (
-            startEntryExists &&
-            hasShiftMovedFromTimecard(actualStartDateTime, actualEndDateTime)
-          ) {
-            setMessages(actualStartDateTime, actualEndDateTime);
-          }
           setTimeEntries(newTimeEntries);
           setSummaryErrors({});
```
joannafawl, line 223: These three lines (`setSummaryErrors({})`, `setShowEditShiftHours(false)` and `setIsErrorVisible(false)`) won't get called if the shift moves from the timecard date, is that something we still want?

Xeena-Newman: after this change, the error messages were persisting if the shift moved from timecard, so I moved setSummaryErrors({}) to the top of the onSubmit, so it fixes that - think this works for setIsErrorVisible too perhaps, since they get cleared

and if the shift moves, then I didn't see why the setShowEditShiftHours(false) needed to be called because it disappears from the timecard anyways, but maybe I'm thinking about that wrong?

joannafawl: Could we delete `setSummaryErrors({})` from here then if it gets called at the top?
I think you're right for `setShowEditShiftHours(false)`, we probably don't need that

Xeena-Newman: yep, that's removed now

joannafawl (approved):

src/components/timecard/edit-shift-hours/EditShiftHours.jsx
```diff
@@ -141,6 +141,7 @@ const EditShiftHours = ({
     dayjs.extend(utc);
 
     setSummaryMessages([]);
+    setSummaryErrors({});
```
IanFisherHO, line 144: Now that the UI review is merged this should be an array here []. Update from main and see if everything still plays ball with this change in.

IanFisherHO (changes requested):

PR #104, opened 2023-05-19
Eahw1624/adding field ids

IanFisherHO:


src/components/annual-target-hours/accruals/AnnualTargetHours.jsx
```diff
@@ -17,6 +17,7 @@ const AnnualTargetHours = ({ targetData, accrualsData }) => {
         <h1 className="govuk-heading-m newline">
           {targetData ? (
             <span
+              data-testid="annualTargetHours-title-span"
```
joannafawl, line 20: How come this one is data-testid instead of id?

Xeena-Newman (approved):

